PR #9, opened 2016-09-08
Install cmt and cmti files

### LINKED ISSUE #7: Build & Install cmt's and cmti's
These are useful for tools such as merlin. For example, to be able to go to definition.

dsheets:
Due to the unconditional use of -bin-annot this patch requires OCaml 4+

Fixes #7.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,8 @@
 *.cma
 *.cmxa
 *.cmxs
+*.cmt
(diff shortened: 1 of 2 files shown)


rgrinberg:
This is no longer needed @mjambon

rgrinberg:
This project now uses jbuilder so this is already done

PR #15, opened 2017-05-16
Bring atdgen

rgrinberg:
While preserving the git history!

if you'd like you can also look into importing issues from the other repo with something like https://github.com/IQAndreas/github-issues-import

I'll do it myself if you give me contributor access.

A couple of other changes I made:

* Removed all LICENSE files but 1. I think 1 license should be enough for everything in this repo.
* Removed the separate version numbers. It probably makes sense to version these together from now on. I'd suggest a version number for the next release that is greater than either of the individual versions so that users upgrade.


mjambon:
Cool, thanks. I added you as a collaborator.

Regarding the licenses: yes, they're the same and I think we should merge them into one. Copyright is about the same too, and anyways it's determined by actual contributions, easy to track since the git history was preserved.

Regarding the version numbers: sure. atdgen is at 1.10.2 and atd is at 1.2.1, so let's start from 1.11.0 which should be compatible with the rules of http://semver.org/.

rgrinberg:
ok @mjambon this is ready. I will update the README in the PR that imports the docs.

mjambon:
ok, let's merge this.

FYI I won't be available until Tuesday due to family visit.

rgrinberg:
NP, this isn't urgent.

PR #16, opened 2017-05-20
Import atdgen documentation

rgrinberg:
@mjambon what do you think about me converting the docs into rst and then hosting it on readthedocs? See http://jbuilder.readthedocs.io/en/latest/ for an example of what it would look like

You'd have to give up on all that css you write, but at least you'll get free search and automatic building of the docs.


rgrinberg:
ping @mjambon :smile:

mjambon:
Moving the docs to readthedocs would be wise. I didn't know about it, it's great.
pandoc might help you with the translation from markdown to rst.

PR #17, opened 2017-05-28
Fix jbuilder errors

SanderSpies:
When I tried to build atd via make I got the following error:

`Error: this field is useless without a (public_names ...) field`

cc @rgrinberg

### DIFF
--- a/atd/test/jbuild
+++ b/atd/test/jbuild
@@ -3,6 +3,7 @@
 (executables
  ((libraries (atd))
   (names (unit_tests))
+  (public_names (unit_tests))  
   (package atd)))
 
 (alias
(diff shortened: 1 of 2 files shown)


SanderSpies:
Even with these "fixes" it doesn't work. Will investigate a bit more. Apologies.

rgrinberg:
Which version of jbuilder are you using? For me that's just a warning.

The problem is that the `package` should be on the aliases i believe rather than the executable stanzas. If you'd like to make a proper fix :)

SanderSpies:
@rgrinberg That's much better! Thanks

SanderSpies:
Yikes, seems like I locally pinned jbuilder and forgot about it - this all just works.

PR #21, opened 2017-08-13
Don't use ocaml syntax just to generate version numbers

rgrinberg:
jbuilder has a ${version:pkg} variable that can be used instead

### DIFF
--- a/atd/src/jbuild
+++ b/atd/src/jbuild
@@ -1,13 +1,3 @@
-(* -*- tuareg -*- *)
-#require "unix"
-
-let version =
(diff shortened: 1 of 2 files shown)


rgrinberg:
Hold off on merging this for now. There's some quirk here that breaks jbuilder's incremental compilation. Investigating...

rgrinberg:
@mjambon ok this is safe to merge. There's a bug in jbuilder that's spoiling incremental compilation whenver's there's an mli for mly/mll files. We'll just omit those for now.

PR #22, opened 2017-08-13
Add a bunch of mli's

rgrinberg:
The types aren't very pretty because they're mostly the inferred types but it's
a decent start. Now we have a clean separation of public/private functions
everywhere which means that we can add docs and start slowly removing dead
code (detectable with jbuilder build --dev)

### DIFF
--- a/atd/src/atd_check.mli
+++ b/atd/src/atd_check.mli
@@ -1,0 +1,1 @@
+val check : [ `Type of Atd_ast.type_def ] list -> unit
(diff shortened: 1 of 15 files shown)


mjambon:
This is great, thanks.

I think we should add headers that explain that these are primarily for internal use and/or subject to change (where applicable).

rgrinberg:
Yeah, definitely. Some of them are definitely unusable for the general user for now. I'll add a warning.

Also, I've made sure that there are no mli's for .mll and .mly files (to workaround a jbuilder bug w/ incremental compilation)

rgrinberg:
I've added some warnings to the mli's with the ugly polymorphic variants. I think the other interfaces are fairly tame and usable and I don't think you had those in mind anyway :)

PR #23, opened 2017-08-16
Massive warnings clean up:

rgrinberg:
The goal of this PR is to build atdgen with:

```
$ jbuilder build --dev
```

I succeeded doing so with the following caveats:

* The commit history looks atrocious. Sorry about that, I'll need to squash
before the merge.

* There are some suspicious looking unused arguments that might be uncovering
potential bugs. Check out the commits with "suspicious" somewhere in the title.
Those are really the only ones that need review.

Important note:

This PR actaully revealed to me some regressions i've introduced in my mli PR.
My mli for the various runtimes was vary conservative and hence doesn't include
many of the functions that are used by generated code. The tests didn't fail at
ony of this, so we really need to improve that. I will fix this regression in an
upcoming PR.

Some other general notes recommedations:

* Polymorphic variants make type errors really really ugly. Is there a reason
why there's so prevalent in this codebase? I'd love to be able to just change
them into normal variants

* The loc arguments to constructors are ignored so much I wonder if it makes
sense to just have 2 ast's with/without locs.

* There are some pretty big tuple arguments to some constructors. We should
really clean those up.

### DIFF
--- a/atdgen/test/github.atd
+++ b/atdgen/test/github.atd
@@ -1,3 +1,23 @@
+(* The following is take from:
+   https://github.com/rgrinberg/ocaml-github/blob/master/lib/github.atd *)
+
+(*
+ * Copyright (c) 2012 Anil Madhavapeddy <anil@recoil.org>
+ * Copyright (c) 2013-2016 David Sheets <sheets@alum.mit.edu>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ *)
 type error = {
   resource: string;
   ?field: string option;
--- a/atdgen/test/github.atd
+++ b/atdgen/test/github.atd
@@ -1,0 +1,1254 @@
+type error = {
+  resource: string;
+  ?field: string option;
+  code: string;
+  ?message: string option;
+} <ocaml field_prefix="error_">
+
+type message = {
+  message: string;
+  ~errors <ocaml default="[]">: error list;
+} <ocaml field_prefix="message_">
+
+type rate = {
+  limit: int;
+  remaining: int;
+  reset: float;
+} <ocaml field_prefix="rate_">
+
+type rate_resources = {
+  core: rate;
+  search: rate;
+} <ocaml field_prefix="rate_resources_">
+
+type rate_limit = {
+  resources: rate_resources;
+} <ocaml field_prefix="rate_limit_">
+
+type scope = [
+  | User <json name="user">
+  | User_email <json name="user:email">
+  | User_follow <json name="user:follow">
+  | Public_repo <json name="public_repo">
+  | Repo <json name="repo">
+  | Repo_deployment <json name="repo_deployment">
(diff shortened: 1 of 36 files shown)


atdgen/test/github.atd
```diff
@@ -0,0 +1,1254 @@
+type error = {
```
mjambon: Could you insert a comment that explains where these definitions come from? (which are impressive, btw)

rgrinberg: Sure, I added a link to the source and  the copyright header (found in the repo itself)

mjambon:
> Polymorphic variants make type errors really really ugly. Is there a reason why there's so prevalent in this codebase? I'd love to be able to just change them into normal variants

Polymorphic reason were chosen over classic ones because I wanted to reuse constructors such as `Int` in different types without scoping issues.

File `a.ml`:
```ocaml
type a = Int | String
```

File `b.ml`:
```ocaml
type b = Int of int | String of string
```

File `c.ml`:
```ocaml
open A
open B

let read (a : A.a) value : B.b =
  match a with
  | Int -> Int (int_of_string value)
  | ...
```

The code above used to fail on the `Int` pattern. The typechecker would complain that it has type `B.b`. Now this works after I requested it to the OCaml core team and they kindly implemented the feature back in 2011 or so. So we could switch to classic variants and it would work nicely, I think.

If it doesn't really solve the problem or for any reason we stick with polymorphic variants, my go-to solution is to add type annotations on function declarations, like I did in the example above. This works for everyday web app development, which don't deal with huge variants and trees like atd, so there may be a specific class of problems that I'm overlooking.

mjambon:
> The loc arguments to constructors are ignored so much I wonder if it makes sense to just have 2 ast's with/without locs.

Do you have a specific example? I mean, wouldn't it introduce complexity and confusion if we have two types instead of one?

> There are some pretty big tuple arguments to some constructors. We should really clean those up.

We have a lot of patterns like `Foo (_, _, _, _, Bar)` but it's not always a good idea to use records instead. Issues to consider include:

* consistency: having tuples or records as arguments of all constructors but not a mix because it would be confusing ("I expected a tuple here, why is there a record?")
* legibility: `Foo (_, _, _, _, Bar)` vs. `Foo { xyz = Bar }` where `xyz` doesn't help much
* extensibility: adding new fields to records breaks less existing code than if we have tuples, but it typically breaks _some_ code anyway
* self-documentation: records are better

Well, that's how I've been thinking about those, for what it's worth. Application code that uses atdgen shouldn't depend on these internals, so we have a lot of flexibility here.

rgrinberg:

> The code above used to fail on the `Int` pattern. The typechecker would complain that it has type `B.b`. Now this works after I requested it to the OCaml core team and they kindly implemented the feature back in 2011 or so. So we could switch to classic variants and it would work nicely, I think.
>
> If it doesn't really solve the problem or for any reason we stick with polymorphic variants, my go-to solution is to add type annotations on function declarations, like I did in the example above. This works for everyday web app development, which don't deal with huge variants and trees like atd, so there may be a specific class of problems that I'm overlooking.

Ok that clarifies things Martin. I always refer to your SO [1] answer
when asked about polymorphic variants, so I'm in full agreement. If we
can get away with normal variants, we should do so. The compiler's
gotten a lot better at letting us do this as you said.

I also don't mind adding type annotations or wrapping types which share
constructor names in modules to help with scoping.

[1] https://stackoverflow.com/a/9367566/1424711

rgrinberg:

> Do you have a specific example? I mean, wouldn't it introduce complexity and confusion if we have two types instead of one?

Sure have a look at how jbuilder handles parsing s-expressions for example:

https://github.com/rgrinberg/jbuilder/blob/include-support-variables/src/sexp.mli#L3

https://github.com/rgrinberg/jbuilder/blob/include-support-variables/src/sexp.mli#L9

You do end up paying a little bit in complexity because you sometimes
have to insert a conversion function that strips off the locs when you
don't care about them. But the benefit is that you have clearer types
for functions that don't care about locs. It's a trade-off to consider,
I'm not insiting on this.

> We have a lot of patterns like `Foo (_, _, _, _, Bar)` but it's not always a good idea to use records instead. Issues to consider include:
>
> * consistency: having tuples or records as arguments of all constructors but not a mix because it would be confusing ("I expected a tuple here, why is there a record?")
> * legibility: `Foo (_, _, _, _, Bar)` vs. `Foo { xyz = Bar }` where `xyz` doesn't help much
> * extensibility: adding new fields to records breaks less existing code than if we have tuples, but it typically breaks _some_ code anyway
> * self-documentation: records are better
>
> Well, that's how I've been thinking about those, for what it's worth. Application code that uses atdgen shouldn't depend on these internals, so we have a lot of flexibility here.

You've brought up a lot of good points. I'll save the opportunities for
records for when they're really obvious (the labels are useful, code is
simplified). I don't think that mixing records and tuples is a bad thing
in itself, after all they both have their strong points. So this change
can be gradual too.

mjambon:
ok, so I'm merging this.

Go ahead if you want to change polymorphic variants into classic ones, assuming it doesn't cause major breakage.

PR #32, opened 2018-02-01
Wrap atdgen

rgrinberg:
This is the first step of making atd namespace its toplevel module names. Since we are using module aliases, we will not incur any penalties like using the old packs (building is still fast, binary size is the same). But this is changing the external a bit. But nevertheless, I don't think many people use the atdgen libraries directly so I think it's a worthy change.


mjambon:
Wow, this is nice.
I also don't think regular users of atdgen would be affected by these changes.

What further changes were you thinking about? We could do the same for the `atd` library. This should only affect developers of tools similar to atdgen and atdj, and the changes are simple.

rgrinberg:
> What further changes were you thinking about? We could do the same for the atd library. This should only affect developers of tools similar to atdgen and atdj, and the changes they would have to make are simple.

Yep. I will do this for atd as well. I agree that this should have very little impact on users. Atdj is in the repository now as well so everything is transitioned is at once.

mjambon:
Thank you!

PR #34, opened 2018-03-14
Small cleanups

rgrinberg:
Just some trivial cleanups

@mjambon I wonder if it makes sense to ocp-indent the whole code base rather than dying it bit by bit?


mjambon:
Great. Thanks.

Yes, go ahead if you want to run ocp-indent on all files. Maybe post the relevant commands in the git commit so that we can apply them to branches in progress and minimize conflicts (there's one old branch on which I might resume work eventually).

rgrinberg:
Ok. Will do so in another pr

Issue #35, opened 2018-03-23
Split runtime into own opam package


rgrinberg:
There's a use case where users would like to vendor the generated code by atdgen into their repos. Unfortunately, this doesn't let you drop the dependency on atdgen as the runtime is still in atdgen itself. It would be nice if this was a separate package to allow this kind of use case. I'm proposing we split out the runtime into a `atdgen-runtime` package for this purpose.

rgrinberg:
Another approach that can be taken here is moving the code generator to its own package. For example leave the runtime as `atdgen` but move the code generator to `atdgen-pp`. This is uglier naming wise, but I think it breaks less code in the short run. I'm thinking of all those people that are using `Atdgen\.` function currently (e.g. validation). But of course we can also we re-export all those symbols in `atdgen`.

mjambon:
I like the `atdgen`/`atdgen-runtime` split and I don't understand how the other proposal would be better (seems worse to me). I'll continue the discussion on the pull request #36 (if needed).

rgrinberg:
This is fixed in master.

PR #36, opened 2018-03-23
Atdgen runtime

rgrinberg:
Okay so this is my attempt at #35. I'm pretty happy with it as it seems to retain full compatibility with old code, and it hides more interfaces from the user. All the code generation stuff is now completely hidden for example. It also lets users specify dependencies more precisely in their opam file:

```
"atdgen" {build}
"atdgen-runtime"
```

And it lets users to eliminate the code generation dependency entirely if they want to vendor the checked in code. In the future, we can even split the runtime into the biniou and yojson components so that users will only use what they generate.


mjambon:
This looks good to me.

A good thing would be to ensure that the preprocessor emits code that's compatible with the runtime. So far this has been guaranteed by installing both from the same source. Does opam provide a way to constrain `atdgen` to the same version as `atdgen-runtime`? If not, perhaps we should use a contraint of the form `DEPENDENCY {>= MY_VERSION}`.

rgrinberg:

> A good thing would be to ensure that the preprocessor emits code
> that's compatible with the runtime. So far this has been guaranteed by
> installing both from the same source. Does opam provide a way to
> constrain `atdgen` to the same version as `atdgen-runtime`?

There's a way to do this in opam 2.0 but IMO this restriction is way too
heavy handed and therefore useless. We should be allowing users to mix
and match different versions of the runtime and the codegen as long as
they are compatible.

> If not, perhaps we should use a contraint of the form `DEPENDENCY {>= MY_VERSION}`.

I'll add a minimum constraint.

rgrinberg:
I added a version constraint. We'll need to release 1.13 for the constraint to make sense. But I suppose that's good anyway.

rgrinberg:
FYI @c-cube you will be able to have a build dependency on `atdgen` following this PR.

mjambon:
Great, thanks!

Let's release 1.13 after #37 is addressed.

PR #38, opened 2018-03-24
Add .descr files

rgrinberg:
This is so that opam submissions use the latest .descr in this repo


rgrinberg:
Okay, so this PR basically adds full topkg support. Try pinning things now and the version number should make more sense.

Let me make the next release and I will some instructions here on how the release process is automated (though it can still be done manually of course).

rgrinberg:
Also, topkg is a bit of stickler for having a proper changelog. I'll keep it maintained if it's a hassle for you.

mjambon:
Hmm, the new version is not picked up by my `opam pin`.

I tried various things, like editing the changelog to change the version from 1.13 to 1.13.0. After that, I ran `topkg tag` which set the correct local git tag. I'm guessing I have to run other `topkg` commands. Which ones?

I didn't have much success with the rest. `opam pin add atd .` installs it as 1.12.0. Same command for `atdgen-runtime` installs it as `~unknown`. Same command for `atdgen` fails because of the version constraint violation (atdgen.opam requires atdgen-runtime >= 1.13.0, which I failed to install properly).

rgrinberg:
> I tried various things, like editing the changelog to change the version from 1.13 to 1.13.0

Ah yes, that's a typo of mine. I pushed a commit to fix to use 1.13.0

> I'm guessing I have to run other topkg commands. Which ones?

No, there are no other commands. You probably have a different expectation of how pinning should work with respect to versions (like the rest of us).

When you say the "new version being picked up", do you mean the version reported by opam after you pin? I certainly don't expect opam to pick up the right version here. When you pin something from the source, opam will simply reuse the same version number as the most recently released version in opam-repository.

What you can do test things locally is:

```
# note that we must specify the version here, or the constraint in the opam file will not work
opam pin add atdgen-runtime.1.13.0 .
```

After this you can pin the rest of the packages with:

```
opam pin add atdgen .
```

etc.

Note that while opam doesn't use 1.13.0, the version reported by `Atd.Version.version` is still correct however.

rgrinberg:
Actually, isn't the constraint on using `atdgen-runtime {>= "1.13.0"}` redundant anyway? It's not like we have an older version of atdgen-runtime. I'll remove it and we can constraints once the code generator breaks compat with some version of the runtime.

rgrinberg:
Okay, so I added relevant constraints on atdj and atdgen. Those must use `atd >= 1.13.0` because they're using the aliased names. You will need to pin atd as follows locally:

```
$ opam pin add atd.1.13.0
```

mjambon:
Pinning works as you instructed, thanks.

However, the version placeholder remains in the installed version:
```ocaml
utop # Atd.Version.version;;
- : string = "%%VERSION%%"
```

mjambon:
Merging! I'll attempt a release with topkg.

rgrinberg:
Thanks. Btw, opam publish is a bit wonky with repos that contain more than
one opam package so you will have to do few manual steps. I can write up a
little makefile target and a small explanation of the workaround, until
opam publish gets proper support. Should be able to do that tonight

mjambon:
That's very nice of you. I don't mind making a release by hand. It would be good to set version constraints on the opam packages that this release breaks. According to `opam list --depends-on=...`, 4 packages list `atd` as a direct dependency and many more list `atdgen` as a dependency. This means around 4 packages are affected, plus or minus of few if dependencies are listed incorrectly.

rgrinberg:
Yup, those constraints can be updated independently. But I'm actually not
sure if we broke anything on opam. The core use case of generating code and
linking in atdgen as the runtime only introduces deprecation warnings. But
of course, it's hard to predict this so the opam ci will let us know.

PR #39, opened 2018-03-28
Add tests that make sure that generated code doesn't change

rgrinberg:
This is a far stronger guarantee that will let us make changes in a much more
controlled fashioned.

I should have added this a while ago, now if we change the codegen we're gonna have to make the tests pass by passing in `--auto-promote` to `$ jbuilder runtest` or running `$ jbuilder promote` following the tests.


rgrinberg:
I'm merging this PR because I'd like the other PR's to show the differences in the generated code. These kinds of tests are already present in cppo, so this is just more of the same. There might be too many overlapping features being tested at the moment, but it's still better than nothing. In any case, it's easy to remove stuff.

PR #41, opened 2018-03-30
Remove a massive amount of polymorphic variants

rgrinberg:
As discussed in #23, polymorphic variants have limited utility in recent versions of OCaml and in fact, I was able to get rid of the vast majority of them with a very minimal amount of type annotations. There's a few polymorphic variants remaining that I will fix soon, but they require a bit more restructuring to do things cleanly (like putting types into their own modules). But the benefit is already there: using merlin now is far easier and all the inferred types are readable.

As for the tuple -> record conversion. I agree that it should probably be done more selectively. I think there's a few examples where the benefit is pretty clear, and I don't think there are issues with mixing records and tuples (things are already mixed). I'll make PR's for those in due time.

### DIFF
--- a/atd/src/ast.ml
+++ b/atd/src/ast.ml
@@ -27,20 +27,19 @@ and module_item =
 and type_param = string list
 
 and type_expr =
-    [ `Sum of loc * variant list * annot
-    | `Record of loc * field list * annot
-    | `Tuple of loc * cell list * annot
-    | `List of loc * type_expr * annot
-    | `Option of loc * type_expr * annot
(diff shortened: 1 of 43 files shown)


rgrinberg:
Btw, I also had to sneak in some indentation changes - it was too hard to keep some of them out. In any case, there's really not much to review here as most of this work was done with search and replacing...

rgrinberg:
Okay, now this PR also shows that none of the generated code for any of the .atd files in the repo changed. So I'm quite certain of the correctness of this PR.

mjambon:
ok, great. Thanks!

PR #44, opened 2018-04-01
[WIP] Support for a Bucklecript-like backend

rgrinberg:
## TODO

- [x] Decoding: runtime
- [x] Decoding: primitive types.
- [x] Decoding: variants
- [x] Decoding: unwrapping
- [x] Decoding: list
- [x] Decoding: records
- [ ] Decoding: arrays
- [x] Decoding: optionals
- [x] Decoding: .mli
- [x] Encoding: runtime
- [x] Encoding: support for primitive types.
- [x] Encoding: support for variants
- [x] Encoding: support for unwrapping
- [x] Encoding: support for list
- [x] Encoding: support for records
- [ ] Encoding: support for arrays
- [x] Encoding: optionals
- [x] Encoding: runtime
- [x] Encoding: mli
- [x] Support for the `-bs` option in the driver.
- [x] Tests: yojson -> bucklescript
- [x] Tests: bucklescript -> yojson
- [x] Support for `-open` flag.
- [ ] Automatically add correct `Open` statements.
- [x] Type annotations: Records
- [x] Type annotations: Variants

## The codec stuff

You can notice that the backend in use isn't being tied to bucklescript in anyway but it is basically fully compatible with [bs-json](https://github.com/glennsl/bs-json). The reason for that is many fold:

* Such a backend can be used for jsoo as well.
* We can't really integrate bucklescript libs into atd b/c of build system incompatibilities.
* I like my own codec abstraction because it can pave the way for generating readers and writers all at once a la https://github.com/OCamlPro/ocplib-json-typed/


mjambon:
I don't want to block you on this. Just let me know if you expect any breaking changes.

rgrinberg:
Nah, I don't plan on breaking anything. While this PR looks quite big, the final version should be a lot smaller. I'd definitely would want you to look over that final one. But all these intermediate steps, probably not worth the time...

rgrinberg:
Okay, so the diff should be more readable now and contain only things that are related to the bucklescript backend. However, the commit history is still a mess. But at least the chances of conflict between this PR and others are now very slim.

I still need to move all the independent test changes to master. Will do that later.

rgrinberg:
@mjambon I'm thinking of just merging this PR and continuing adding support in separate PR's. The overhead of keeping this PR is rebased is non trivial, and the feature itself is completely isolated from users. If they don't use bucklescript, it's not really going to impact them.

mjambon:
I merged this.

I had a very quick glance at the code and I was wondering what the new `atdgen-codec-runtime` was and how it relates to `atdgen-runtime`. I think we should describe the project structure in `internal-doc/` so as to help contributors figure out what's relevant to them.

PR #46, opened 2018-04-03
re-indent oj_emit

rgrinberg:
This will make the buckle PR a little cleaner.


rgrinberg:
I'm merging as there's nothing interesting here. One can double check this on their own by ignoring the whitespace differences: https://github.com/mjambon/atd/pull/46?w=1

PR #47, opened 2018-04-03
remove all opens of Error

rgrinberg:
It's quite a bit easier to grep for failwith with the fully qualified names. I'll also rename `Error.error` to something like `Error.fatal` to be a bit more descriptive.


mjambon:
great, thanks.

PR #48, opened 2018-04-03
Remove various aliases

rgrinberg:
These don't seem too useful, let's reduce the numbe of identifiers.

### DIFF
--- a/atdgen/src/ox_emit.mli
+++ b/atdgen/src/ox_emit.mli
@@ -37,17 +37,17 @@ val write_ocaml : target -> string -> string -> unit
 
 val name_of_var : string -> string
 
-val nth : Mapping.loc_id -> int -> int -> Mapping.loc_id
+val nth : string -> int -> int -> string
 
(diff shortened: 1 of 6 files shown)


rgrinberg:
Fairly harmless change that doesn't require much scrutiny.

atdgen/src/ox_emit.mli
```diff
@@ -37,17 +37,17 @@ val write_ocaml : target -> string -> string -> unit
 
 val name_of_var : string -> string
 
-val nth : Mapping.loc_id -> int -> int -> Mapping.loc_id
+val nth : string -> int -> int -> string
```
mjambon: I was wondering why you didn't like having a type alias `loc_id = string`. The original intent was to document function signatures without the burdening ourselves with an abstract type.

rgrinberg:
It's not that I didn't like it, it's just that i don't think it was used
consistently. I was thinking of reintroducing this but with a private alias
so that we can use it more consistently

PR #69, opened 2018-04-20
Remove obsolete checks for overlaps

### LINKED ISSUE #53: Remove -o-name-overlap and -o-no-name-overlap options
As far as I understand, these options are only useful for obsolete versions of OCaml: < 4.01. It's doubtful that atd even works on those versions since it doesn't even build. Dune only works on >= 4.02 anyway.

rgrinberg:
These checks are only useful for versions of OCaml taht atdgen doesn't even
build on (< 4.02).

Fix #53

### DIFF
--- a/atdgen/bin/ag_main.ml
+++ b/atdgen/bin/ag_main.ml
@@ -49,12 +49,6 @@ let parse_ocaml_version () =
   else
     None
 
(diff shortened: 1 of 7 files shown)


mjambon:
Looks good to me.

PR #72, opened 2018-04-23
json adapters

mjambon:
Still working on it. Feature was implemented and basic test passes.

I'll leave comments next to the interesting changes if you want to take an early peek.

At least I still have to fix merge conflicts and emulate the tag_field feature which I removed.

I probably won't touch this again until next weekend (Apr 28-29). Let's minimize changes in `oj_emit.ml` in the meantime.


atdgen-runtime/src/oj_run.ml
```diff
@@ -219,6 +219,20 @@ let missing_fields p bit_fields field_names =
...
+  let lb' = Lexing.from_string s' in
+  reader p lb'
+
+let write_with_adapter restore writer ob x =
+  let ob_tmp = Bi_outbuf.create 1024 in
+  writer ob_tmp x;
+  let s_tmp = Bi_outbuf.contents ob_tmp in
+  let ast = Yojson.Safe.from_string s_tmp in
+  let ast' = restore ast in
+  Yojson.Safe.to_outbuf ob ast'
```
mjambon: These two runtime functions apply the user-defined functions `normalize` and `restore` to the json AST. This comes with the cost of an extra read+write.

The big advantage for the user is that `ast` and `ast` have the same, generic type which allows the `normalize` and `restore` functions to be defined independently from the user's atd types.

atdgen/src/json.ml
```diff
@@ -33,7 +37,7 @@ type json_repr =
   | Option
   | Record of json_record
   | String
-  | Sum
+  | Sum of json_adapter option
```
mjambon: At this stage, only atd variant types support the `<json adapter=...>` annotation.

It may be useful to support it on records as well; see https://github.com/mjambon/atd/issues/57 for an example.

atdgen/src/json.ml
```diff
@@ -78,13 +124,14 @@ let get_json_fname default an =
...
   Atd.Annot.get_flag ["json"] "keep_nulls" an
 
 let get_json_record an =
   {
     json_keep_nulls = get_json_keep_nulls an;
   }
+
+let tests = [
+  "parse adapter name", test_parse_adapter_name;
+]
```
mjambon: to do: run this test suite with the other test.

atdgen/test/adapter.ml
```diff
@@ -0,0 +1,29 @@
...
+
+  (* Add a `"type"` field. *)
+  let restore (x : json) : json =
+    let cons, assoc =
+      match x with
+      | `List [ `String cons; `Assoc l ] -> cons, l
+      | _ -> assert false
+    in
+    `Assoc (("type", `String (String.lowercase_ascii cons)) :: assoc)
+end
```
mjambon: This is only used for a test at the moment but I want to add this and more patterns to the atdgen runtime that users could use without hassle.

atdgen/test/test3j_t.expected.mli
```diff
@@ -0,0 +1,26 @@
...
+  patch1: int option option;
+  patch2: int option option;
+  patch3: int option option
+}
+
+type b = { thing: int }
+
+type a = { thing: string; other_thing: bool }
+
+type adapted = [ `A of a | `B of b ]
```
mjambon: The 4 `test3j_[tj].ml*` files were not checked in, presumably because they were git-ignored.

atdgen/src/oj_emit.ml
```diff
@@ -35,6 +35,7 @@ type param = {
 
 }
 
+let unopt = function None -> assert false | Some x -> x
```
rgrinberg: This is `Option.value_exn` in Import btw.

atdgen/src/oj_emit.ml
```diff
@@ -554,89 +439,48 @@ let rec make_writer p (x : Oj_mapping.t) : Indent.t list =
...
+          Line (sprintf "Bi_outbuf.add_char ob %C" cl);
+        ]
       ]
 
 and make_record_writer p a record_kind =
+  let dot =
+    match record_kind with
+        Record -> "."
+      | Object -> "#"
+  in
```
rgrinberg: You can just use `Ocaml.dot` here.

rgrinberg:
Just to confirm, we aren't breaking the github library in any way here? It's a fairly important consumer of this library.

mjambon:
I am breaking the GitHub library at the moment. The Type_and_value_fields
submodule will help with porting, but it requires a functor application for
each new pair of type/value fields. I haven't tested it yet. I'm hesitating
to preserve syntactic support for the tag_field annotation.

atdgen/src/oj_emit.ml
mjambon: Fixed, thanks. I might have brought this old code back when undoing the tag_field implementation.

atdgen/src/oj_emit.ml
mjambon: Fixed.

atdgen/test/test_atdgen_main.ml
```diff
@@ -955,41 +584,29 @@ let all_tests = [
...
+(* TODO: use Alcotest to run the test suite. *)
 let quality_test () =
   List.iter (fun f ->
     try f ()
     with Failed -> ())
     all_tests;
   match List.rev !errors with
-  | [] -> ()
+  | [] ->
+      printf "Success.\n%!"
```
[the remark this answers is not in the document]
mjambon: Makes sense. Reverted.

ocamlflags.sexp
```diff
@@ -1 +1,10 @@
...
\ No newline at end of file
+(:standard
+
+  ; Compiler errors and warnings
+  -w -40-43+32+33        ; default set of warnings -this +that
+  -safe-string
+  -short-paths
+  -strict-sequence
+  -strict-formats
+  -keep-locs
```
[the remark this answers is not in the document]
mjambon: I'm not sure anymore how I came up with this. I simplified the flags and indicated my intent in the file.

atdgen/test/test3j.atd
```diff
@@ -14,122 +14,31 @@ type unixtime_list = int <ocaml repr="float"> list
...
-type fallback_constr = [
-| A <json name="a">
-| Other <json untyped> of (string * json option)
-]
+   See normalize/restore functions in `adapter.ml`.
+*)
+type adapted = [
+  | A <json name="a"> of a
+  | B <json name="b"> of b
+] <json adapter.ocaml="Atdgen_runtime.Util.Json.Adapter.Type_field.Default">
```
mjambon: ^^^ this is how the new feature is used. We can now have dots in annotation field names. I'm worried about the length of this module path. Is this bad? acceptable?

mjambon:
I  made the required [changes to ocaml-github](https://github.com/mjambon/ocaml-github/tree/atd-adapter) so it works when `tag_field` goes away.

I still need to add back support for `<json untyped>` which is also used a lot in `github.atd`. I can't think of a good alternative. It should be straightforward to implement, though; not sure why I removed it the other day.

mjambon:
After a bit of pondering, I think it would be good reimplement only part of the `<json untyped>` feature. In the case of a pure enum&em;the most common case&em; we had this:
```ocaml
type t = [
  | A
  | Bc
  | Other <json untyped> of (string * json option)
]
```
Now instead we'd write this:
```ocaml
type t = [
  | A
  | Bc
  | Other of string
] <json open_enum>
```
This requires the json elements to be simple strings, and those that are neither `"A"` nor `"Bc"` end up under `Other`. An important difference is that the atd tree is no longer dependent on the `json` type (the argument of `Other` being now just a string).

This feature only covers enums. Variants with arguments would need to be dealt with by writing a custom adapter. This is a bit of work, but such situations are expected to be less common.

rgrinberg:

> +type adapted = [
> +  | A <json name="a"> of a
> +  | B <json name="b"> of b
> +] <json adapter.ocaml="Atdgen_runtime.Util.Json.Adapter.Type_field.Default">

The Util module isn't particularly descriptive, so if there's an
opportunity for simplification here, I think it would start by splitting
that module. But otherwise, this looks fine.

mjambon:
@rgrinberg 
`<json adapter.ocaml="Atdgen_runtime.Util.Json.Adapter.Type_field.Default">` is now `<json adapter.ocaml="Atdgen_runtime.Json_adapter.Type_field">`.

mjambon:
@rgrinberg I think we can merge this. Since last time you posted here, I added the `<json open_enum>` feature. I also made changes to ocaml-github to make it work again.

rgrinberg:
Awesome. Thanks. Give me a day to give it one more read and I will give
this a merge if you don't mind.

Makefile
```diff
@@ -1,10 +1,10 @@
 JBUILDER ?= jbuilder
 
 all:
-	$(JBUILDER) build --dev
+	$(JBUILDER) build
 
 tests:
-	$(JBUILDER) runtest --dev
+	$(JBUILDER) runtest
```
rgrinberg: Why did you remove `--dev` here? You don't like some of the warnings?

atdgen/src/json.ml
```diff
@@ -62,9 +86,38 @@ let get_json_float an : json_float =
...
+   --> { ocaml_adapter = Some "Foo.Bar";
+         java_adapter = None; }
+*)
+let get_json_adapter an =
+  let ocaml_adapter =
+    Atd.Annot.get_field (fun s -> Some (Some s))
+      None ["json"] "adapter.ocaml" an
+  in
+  let java_adapter =
+    Atd.Annot.get_field (fun s -> Some (Some s))
```
rgrinberg: Can you remind me why we need the double option type here (and other places)?

Makefile
mjambon: Yes. I remember being annoyed by warning 27. There may be others.
> 27 Innocuous unused variable: unused variable that is not bound with let nor as, and doesn’t start with an underscore (_) character.

I want to write patterns like `Node (loc, x, _)` without having to write `_loc` in case I don't use `loc` later. I don't want to write `Node (_, x, _)` either because the day I need to produce an error that uses `loc`, I don't know where to find it. I also don't want in the same pattern-matching to haves cases like `Node (_loc, x, _)` and other cases like `Other (loc, y)`.

atdgen/src/json.ml
mjambon: I know it's difficult to make sense of it. I'll try to clarify this with commented examples.

atdgen/src/oj_emit.ml
```diff
@@ -124,120 +118,44 @@ let get_assoc_type deref loc x =
...
-  a
+  List.map (
+    fun x ->
+      let ocaml_fname, ocaml_default, json_fname, optional, unwrapped =
+        match x.f_arepr, x.f_brepr with
+        | Ocaml.Repr.Field o, Json.Field j ->
+            let ocaml_default =
+              match x.f_kind with
+                With_default ->
+                  (match o.Ocaml.ocaml_default with
```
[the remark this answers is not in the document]
mjambon: It's very likely because I undid features by restoring old code from before those features. I'll take a look at your cleanup commits and try to re-apply them.

rgrinberg:
I see. The issue is that that this warning is quite useful in other
situations. There's a way to turn it off selectively with annotations, but
I think then we lose portability with 4.02. I guess we can ignore it for
now.

atdgen/src/oj_emit.ml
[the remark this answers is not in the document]
mjambon: ok, I hand-restored your improvements to the `get_fields` function.

atdgen/src/json.ml
mjambon: I created `Atd.Annot.get_opt_field` and added labeled arguments for clarity.

Issue #116, opened 2018-05-30
Fix link formatting and broken links in new docs


mjambon:
e.g. http://atd.readthedocs.io/en/latest/tutorial.html

This is done by editing the source documents under `/doc`. The following should work:

1. Fork the project.
2. Edit documents under `/doc` from the github web interface.
3. Save and check github's preview.
4. When satisfied, make a pull request.
5. Once merged into master, the changes will automatically appear on http://atd.readthedocs.io

PR #117, opened 2018-05-31
fix link formats

anuragsoni:
Relates to #116


mjambon:
Awesome, thank you! Let's see how it looks on http://atd.readthedocs.io/en/latest/

PR #126, opened 2018-06-03
Simplify functions that take f, x and immediately f x

rgrinberg:
It's simpler to just to pass (f x)


rgrinberg:
I've added the last commit because get_fields is useful for the bucklescript branch  as well. Also, 6 elements is just about the limit for tuples 😆

mjambon:
This is good. Thanks.

Issue #33, opened 2018-03-07
Better validator error messages?


brendanlong:
I was trying to use atdgen to validate fields in a record, similar to the example in the docs:

```
type t =
  { id : string
  ; start_date : date
  ; ?end_date : date option }
  <ocaml valid="Requisition_util.validate_requisition">
```

```
(* requisition_util.ml *)
let validate_requisition { Requisition_t.start_date ; end_date } =
  match end_date with
  | None -> true
  | Some end_date -> Date.(end_date >= start_date)
```

The problem is that the error message is just:

> Validation error; path = &lt;root>

I looked at the code in atdgen/src/validate.ml, and it doesn't seem like there's any way to set `msg`. Is there some way to make the message more useful? I'd really like to return something like "end_date must be >= to start_date".

mjambon:
What you can do is use `validator=` instead of `valid=`. This lets you specify an error message as part of the return value of the validator. The type of your validator function should be `Ag_util.Validation.path -> <your type> -> Ag_util.Validation.error option`. `Ag_util.Validation` provides functions to somewhat help create these validators, but it's cumbersome and it's not documented as it should (see https://github.com/mjambon/atdgen/blob/master/src/ag_util.mli).

mjambon:
Perhaps we could improve the situation by letting the user raise an exception from their validator, and the generated code would catch it and use it as an error message. Your particular example would be written as:

```ocaml
(* requisition_util.ml *)
let validate_requisition { Requisition_t.start_date ; end_date } =
  match end_date with
  | None -> true
  | Some end_date ->
      Date.(end_date >= start_date)
      || failwith "end_date must be >= to start_date"
```

brendanlong:
Ah cool, using `validator=` works for now, but the exception version would be more user-friendly if you feel like making that change :)

shonfeder:
I'd be interested in helping with this, if no one is working on it!

I haven't looked at the code yet, but what if we did something like the following (to add additional "just works" functionality)?

1. If an exception is raised by the validation function, catch the message and use that (as @mjambon described).
2. If no error is raised, just supply the name of the validation function for message

Since this case fit (2), @brendanlong would have then seen something like

```
Validation error; expected "Requisition_util.validate_requisition"
```

mjambon:
[...]

PR #132, opened 2018-06-19
Include name of validation function in error message from `valid`

shonfeder:
Aims to address part of #33 

Currently, adding validation to types with the `valid` field produces error messages with no indication of what cased the failure (beyond the path). This change ensures that all validation errors include an indication of the failing validation function. This is the default behavior and messaging I would expect as a user.

E.g., 

Given

```ocaml
(* scratch_util.ml *)

let validate_u8 i = 0 <= i && i <= 255
```

and 

```ocaml
(* scratch.adt *)

type u8 = int <ocaml valid="Scratch_util.validate_u8">

type point = {x: u8; y: u8}
```

We get validation errors like so:

```
Validation error: Failed check by Scratch_util.validate_u8; path = <root>.y
```

Some points to note:

1. This doesn't fully satisfy #33, since the reporter wanted a light-weight way of specifying a particular error message, not just recording the validation function.
2. The formatting of the generated ocaml isn't very nice. I could use pointers to any preferred style guidelines or principles to follow for fixing that up.
3. This PR should include updates to the documentation, which I'll add if the approach is judged to be agreeable.
4. Would it be more fitting to have the error message read `Validation error: validator = Scratch_util.validate_u8; path = <root>.y`?

### DIFF
--- a/atdgen/src/validate.ml
+++ b/atdgen/src/validate.ml
@@ -24,10 +24,13 @@ type validate_repr = (string option * bool)
 
 let make_full_validator s =
   sprintf "\
-    fun path x -> \
-      if ( %s ) x then None \
-      else Some (Atdgen_runtime.Util.Validation.error path)"
-    s
+  fun path x ->
+    let msg = \"Failed check by %s\" in
+    if (%s) x then
+      None
+    else
+      Some (Atdgen_runtime.Util.Validation.error ~msg path)"
+    (String.escaped s) s
 
 let get_validator an =
   let full =
--- a/atdgen/test/test.atd
+++ b/atdgen/test/test.atd
@@ -208,3 +208,5 @@ type validate_me =
     <ocaml valid="fun s -> true">
       list
         <ocaml valid="fun l -> true">
+
+type validated_string_check = string <ocaml valid="fun s -> s = \"abc\"">
--- a/atdgen/test/test.expected.ml
+++ b/atdgen/test/test.expected.ml
@@ -18,6 +18,8 @@ type p = [ `A | `B of r | `C ]
 
 and r = { a: int; mutable b: bool; c: p }
 
+type validated_string_check = string
+
 type validate_me = string list
 
 type val1 = { val1_x: int }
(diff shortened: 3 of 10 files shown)


mjambon:
Great. This is already a net improvement over what we already have.

1: we can merge this branch since it's done and non-controversial. #33 where the user specifies a custom error message can be done later.

2: the formatting of the generated looks good enough to me. The requirements are that it's as easy to debug as possible. It's not meant to be read frequently by humans, so it's fine if it's not perfectly formatted.

If we want to nitpick, I'd say I'm not a fan of the whitespace padding within parentheses. We're not using it elsewhere, so you can remove it. AFAIK, the only risk is if the user inserts code that starts or ends with a `*`&mdash;which is illegal anyway&mdash;in which case that would start a comment and produce weird error messages. Otherwise, no operators combine with `(` or `)` so it's safe enough to write code like `"(" ^ user_input ^ ")"`.

3: no need for documentation if this PR stops here. It would be needed for the next one which supports custom error messages.

4: I'd rather have error messages that look more like plain English. There's no need to make the error message particularly terse. Also, I think the current `path = <root>.y` is confusing. Perhaps something like "field y" would be good; "field [2].x.y" would replace "path = <root>[2].x.y". I'll let you think about it. This also can be done in another PR.

mjambon:
ok, I'm merging this. Thanks :+1: 
Keep the PRs coming :-)

shonfeder:
Thanks for the detailed feedback, @mjambon! The information on general conventions/principles is very helpful.

I'm happy to tweak the formatting to remove the padding around the parens. If this still isn't merged by the time I get home tonight, I'll make a quick fix for that. But I guess this is a quite minor detail in any case.

mjambon:
oh, I thought I had merged it. I'll let you make your changes then.

shonfeder:
I finally made it home! I also made the discussed changes :)

Thanks again for the responses and feedback. I'll follow up on all the things discussed, one way or another.

atdgen/src/validate.ml
```diff
@@ -24,10 +24,13 @@ type validate_repr = (string option * bool)
...
-      if ( %s ) x then None \
-      else Some (Atdgen_runtime.Util.Validation.error path)"
-    s
+  fun path x ->
+    let msg = \"Failed check by %s\" in
+    if (%s) x then
+      None
+    else
+      Some (Atdgen_runtime.Util.Validation.error ~msg path)"
+    s s
```
mjambon: The first `%s` is in a string literal, so I think it should quoted. Check what you're getting if the argument of `make_full_validator` contains double quotes, like `"(fun s -> s = \"abc\")"`.

mjambon: Lokks good to me other than that.

shonfeder: That's good thinking. I'll add some test cases for that.

atdgen/src/validate.ml
```diff
@@ -22,12 +22,34 @@ type validate_repr = (string option * bool)
...
+  let increase_escaped_quote_level = function
+    | Str.Text s  -> s
+    | Str.Delim s ->
+        let escapes_needed = String.length s in
+        let new_escapes = String.init escapes_needed (fun _ -> '\\') in
+        (new_escapes ^ s)
+  in
+  Str.full_split escaped_quote s
+  |> List.map increase_escaped_quote_level
+  |> String.concat ""
```
mjambon: oh you didn't have to go through that trouble. OCaml has this built-in: `String.escaped` for escaping string content without the enclosing double-quotes (`ab"c` -> `ab\"c`) or `sprintf "%S"` for creating a full quoted string literal (`ab"c` -> `"ab\"c").

shonfeder: Oh doh! My initial searches didn't turn up anything, but apparently I didn't look well enough. Thanks!

shonfeder:
Fixed! :)

mjambon:
Looks good. Thank you!

shonfeder:
Thanks for your help, review, and patience!

PR #133, opened 2018-06-19
eta-expand functions using json adapters

mjambon:
When adding code generation for the json wrappers, I didn't remember correctly what the `"fun"` annotations were for. This had the consequence of introducing code like this:
```ocaml
let rec f = wrap g
and h x = x
```
The above is rejected by the ocaml compiler because `wrap g` involves a computation (or something along these lines). Instead the `let f =` must guarantee that it doesn't compute anything. This is achieved by eta-expansion as follows:
```ocaml
let rec f x = wrap g x
and h x = x
```
We use annotations in the generated code (`Annot` nodes) to indicate that eta-expansion is *not* needed.

### DIFF
--- a/atdgen/src/oj_emit.ml
+++ b/atdgen/src/oj_emit.ml
@@ -256,9 +256,9 @@ let write_with_adapter adapter writer =
       let restore =
         Oj_mapping.json_restorer_of_adapter_path adapter_path in
       [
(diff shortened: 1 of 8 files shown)


mjambon:
This fixes bug https://github.com/mjambon/atd/issues/131

PR #137, opened 2018-06-24
Merge atdj documentation

### LINKED ISSUE #109: Consolidate documentation - mechanical merge of legacy docs
@rgrinberg already set up http://atd.readthedocs.io
If I'm not mistaken, so far it contains the atdgen tutorial and the atd syntax reference. We still need to import the atdgen reference manual and polish thing up. I'm not sure how to include ocamldoc- or odoc-generated documentation.

Ideally, we want all the documents listed at https://mjambon.github.io/atdgen-doc/ to be incorporated into http://atd.readthedocs.io

mjambon:
I translated the old latex document with pandoc and made some minimal cleanup. This closes #109.

You can preview it here: https://atd.readthedocs.io/en/atdj-doc/atdj.html (I made it "Protected" so it doesn't show up in the public listing; let me know if you can't access it or if it requires annoying login)


rgrinberg:
What's wrong with having in the main listing however? Seems like atdj is a good tool that should get a little more visibility.

mjambon:
Nothing wrong with that. No, I was referring to the list of the different versions of the atd documentation. We have "latest" and "stable". Here I added the branch `atdj-doc` for preview purposes only (Log in -> Settings -> Versions). This creates a version of the documentation called "atdj-doc". My goal was to make sure random visitors don't accidentally see this document.

rgrinberg:
Ah yes, I see what you mean. I can see the docs just fine btw.

doc/atdj.rst
```diff
@@ -0,0 +1,319 @@
...
+========================
+
+In this section we describe the process of generating a Java interface
+from an ATD specification.
+
+A Java interface is generated from an ATD file as
+
+::
+
+    ./atdj -package <package> <atd_file>
```
rgrinberg: A user is unlikely to have atdj in their current working directory, so i'd change this to `atdj` instead.

mjambon: fixed that.

doc/tutorial.rst
```diff
@@ -1131,7 +1131,7 @@ and ``.ml`` produced by ``atdgen``.
 GNU Make
 --------
 
-We provide `\`Atdgen.mk\` <https://github.com/mjambon/atdgen-make>`__, a generic
+We provide `Atdgen.mk <https://github.com/mjambon/atdgen-make>`__, a generic
```
rgrinberg: Should we just move this makefile into atd as well? It's quite easy to install into same shared directory with jbuild and it will let people reuse it with `-include $(opam config var share)/Atd.mk`.

mjambon: Yes, it's a good idea. I'm running out of time for this week. I'm creating an issue for it.

mjambon: -> https://github.com/mjambon/atd/issues/139

PR #138, opened 2018-06-24
Cleanup assoc and option handling

rgrinberg:
I promise I'm not planning to start crusading against exceptions in favor of result and optional but this is one case where it's worth doing. `Not_found` is a real pain in the neck to debug, because so many things throw it and it doesn't tell you anything about what wasn't found. We end up converting `Not_found` to none in many cases anyway, so let's just take that as the default. The old `assoc` is still available as `assoc_exn` in the few odd cases where it's still necessary.


mjambon:
Yes, this is good. Thanks.

PR #141, opened 2018-07-03
Update data validation tutorial section and example code

### LINKED ISSUE #134: Clarify documentation
@shonfeder wrote in https://github.com/mjambon/atd/issues/33#issuecomment-398278382

> I think the documentation for validation may be out of date? (The docs say valid annotations on a type t should result in validators of the type `t -> bool`, but it looks like validators are actually of type `path -> t -> Validation.error option`). If this correct, I can make a PR to update the docs.

shonfeder:
Closes  #134 

The whitespace changes to doc/tutorial-data/validate/resume_util.ml are just from bringing the example code up to date with the code in the tutorial text.

### DIFF
--- a/doc/tutorial.rst
+++ b/doc/tutorial.rst
@@ -546,38 +546,79 @@ example is:
 
   type t = string <ocaml valid="fun s -> String.length s >= 8"> option
 
-``atdgen -v`` will produce something equivalent to the following
-implementation:
+As we can see from this example, the validation function is specified using the
+annotation ``<ocaml valid="p">``, where ``p`` is a predicate ``p : t -> bool``,
+returning ``true`` when the value of type ``t`` is valid and ``false``
+otherwise.
+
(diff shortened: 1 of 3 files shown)


doc/tutorial.rst
```diff
@@ -546,38 +546,79 @@ example is:
 
   type t = string <ocaml valid="fun s -> String.length s >= 8"> option
 
-``atdgen -v`` will produce something equivalent to the following
-implementation:
+As we can seefrom this example, the validation function is specified using the
+annotation ``<ocaml valid=p>``, where ``p`` is a predicate ``p : t -> bool``,
+returning ``true`` when the value of type ``t`` is valid and ``false``
+otherwise.
```
mjambon: 2 things here:
* seefrom -> see from
* `<ocaml valid=p>` would be clearer as `<ocaml valid="p">` I think.

shonfeder:
Thanks for the corrections! Changes made.

Issue #92, opened 2018-05-27
Get rid of Obj.magic for record readers


mjambon:
_From @mjambon on June 18, 2016 9:51_

This is a followup to https://github.com/mjambon/atdgen/issues/45 and https://github.com/mjambon/atdgen/pull/46.

Get rid of Obj.magic for record readers and instead, use plain old `option ref`s:

``` ocaml
let x = ref None in
let y = ref None in
...
| "x" -> x := Some (read_x_type ...)
| "y" -> y := Some (read_y_type ...)
...
{
  (* optional field *)
  x = !x;
  (* required field *)
  y = (match !y with
      | None -> ... (* error *)
      | Some v -> v
}
```

Perhaps fields with a precomputed default could use a `ref` instead of an `option ref`:

``` ocaml
let y = ref the_default in
...
| "y" -> y := read_y_type ...
...
```

but only if `the_default` is a precomputed value, since we don't want to compute a default value for nothing.

_Copied from original issue: mjambon/atdgen#51_

mjambon:
I did much of that work in an branch that's no longer mergeable: https://github.com/mjambon/atdgen/tree/mj/no-magic
Since then we got rid of some code that was standing in our way (with atd > 1.13).

mjambon:
For motivation see bug https://github.com/mjambon/atd/issues/96

mjambon:
At this time, `Obj.magic` is no longer used in ocaml code for reading json (https://github.com/mjambon/atd/pull/143) but it's still present in code dealing with the biniou format.

mjambon:
Duplicate. Keeping #147.

PR #143, opened 2018-07-14
Generate ocaml/json code without Obj.magic

mjambon:
With this, reading json objects no longer relies on `Obj.magic` (#92 ). Json support is now done entirely without the use `Obj`, which will give a peace of mind to users and maintainers alike.

The biniou/ocaml backend still uses Obj.magic in the same place (reading one record) in addition to initializing and reading its special "table" construct which is a list of records optimized for compactness. I'm reluctant to touch this right now due to the lack of benchmarks and lack of time on my end, and given that the primary goal of biniou was to have the flexibility of json with increased performance.

### DIFF
--- a/atdgen-runtime/src/oj_run.ml
+++ b/atdgen-runtime/src/oj_run.ml
@@ -203,6 +203,8 @@ let missing_tuple_fields p len req_fields =
            (if List.length missing > 1 then "s" else "")
            (String.concat ", " (List.map string_of_int missing)))
 
+let missing_field p field_name =
+  error_with_line p (sprintf "Missing record field %s" field_name)
 
 let missing_fields p bit_fields field_names =
   let acc = ref [] in
--- a/atdgen-runtime/src/oj_run.mli
+++ b/atdgen-runtime/src/oj_run.mli
@@ -36,6 +36,7 @@ val read_number : float read
 val invalid_variant_tag : Yojson.Lexer_state.t -> string -> _
 
 val missing_tuple_fields : Yojson.lexer_state -> int -> int list -> _
+val missing_field : Yojson.lexer_state -> string -> _
 val missing_fields : Yojson.lexer_state -> int array -> string array -> _
 
 val write_with_adapter :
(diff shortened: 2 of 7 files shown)


rgrinberg:
Btw, since this PR is likely restoring compat for older OCaml, let's check that atd now works everywhere where we expect it to work. We can do this by:

* Merging #144 
* Rebasing this PR on master
* running `$ make all-supported-ocaml-versions` (assuming that you have all the switches setup)
* updating jbuild-workspace.dev to reflect the new support matrix (assuming that we can restore 4.02.3 and 4.03.0 compat)

mjambon:
The OCaml code that reads the biniou format still uses `Obj.magic`, together with `Sys.opaque_identity` if the version of OCaml permits. These are the relevant lines:
https://github.com/mjambon/atd/blob/remove-obj-magic/atdgen/src/ob_emit.ml#L658-L662

That said, I think we do support ocaml < 4.03 because `Sys.opaque_identity` is used only if the ocaml version is >= 4.03. It's fine because it fixes a problem that only occurred starting with 4.03, according to the [detailed description](https://github.com/mjambon/atd/blob/remove-obj-magic/atdgen/test/test_atdgen_main.ml#L60-L174) left by @gasche. A problem for us is that the expected output (based on ocaml >= 4.03) doesn't match the actual output when using ocaml < 4.03, so the expect tests will fail for these earlier versions of ocaml. Do you know a way to deal with this?

mjambon:
I must say I don't know who uses biniou at the moment. It may not be very many people. I submitted a poll here: https://discuss.ocaml.org/t/input-wanted-from-biniou-users-binary-format-supported-by-atdgen/2300

Ideally I'd like to remove all `Obj.magic` from the atd codebase without conducting any benchmark, or let interested people conduct them whenever they get a chance.

gasche:
[Vecosek](https://gitlab.com/smondet/vecosek) uses Biniou for performance reason, so @smondet may be able to give feedback on the performance impact of a change.

rgrinberg:
Since this doesn't seem like a problem for @smondet, can we merge this? Martin, anything else left to do here?

mjambon:
@rgrinberg this branch can be merged and won't break anything. I'll create a separate issue for removing Obj.magic from biniou since it's a bit more work and I don't have the time to do it at the moment.

PR #150, opened 2018-08-28
Generate typedefs by default + respect -nfd and -ntd cli flags

cyberhuman:
This PR makes `atdgen -bs` output type declarations by default, since:
1. This is what all other generators (`-j`, `-b`, `-v`) do;
2. It is more convenient because `-open Foo_t` is not needed.

It is still possible to get the old behaviour by passing `-ntd` to atdgen (but why?).

In addition, `-nfd` is now supported, too.

### DIFF
--- a/atdgen/src/obuckle_emit.ml
+++ b/atdgen/src/obuckle_emit.ml
@@ -55,7 +55,7 @@ let destruct_sum (x : Oj_mapping.t) =
   | _ -> Error.error (loc_of_mapping x) "Cannot destruct unknown type"
 
 
-let make_ocaml_bs_intf buf _deref defs =
+let make_ocaml_bs_intf ~with_create buf deref defs =
   List.concat_map snd defs
   |> List.filter Ox_emit.include_intf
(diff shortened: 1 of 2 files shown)


cyberhuman:
ping

mjambon:
I'll let @rgrinberg review this as he's the author of the bucklescript backend.

Rudi, if there's anything we can do to speed up code review and approval, please suggest. For example, we could improve the guidelines and criteria for submissions. Right now there's only a generic external document mentioned in the main readme (https://github.com/mjambon/documents/blob/master/how-to-contribute.md). If we could have a shorter version, more specific to atd that might be a good thing.

rgrinberg:
Assigning reviewers like you did should be enough. Sorry about the delay this time, the patch itself is straightforward.

rgrinberg:
> It is still possible to get the old behaviour by passing -ntd to atdgen (but why?).

As we've discussed before, to me this seems like the better default. Why generate so much extra code when a single `open` will do? In any case, I agree that the backends should be consistent.

Issue #130, opened 2018-06-15
[feature request] allow adapters in more locations


ygrek:
Usecase:
Some api returns a field in record sometimes as string and sometimes as string list. This can be currently hacked around with `json wrap` but looks like adapter could provide a natural solution when applied on record field position.

gtrak:
I could use this on plain lists, was confused that it didn't work given the documentation

mjambon:
@gtrak I agree with the need to support the feature on lists. I actually thought we did so it's possible that I wrote that somewhere (in an announcement maybe?). The reference manual is correct, though (https://atd.readthedocs.io/en/latest/atdgen.html#field-adapter-ocaml).

gtrak:
I wasn't able to make it work, then traced it to the source and the emitter seems specific to record and variant types?  If we don't yet support it, it might be better to syntax error than silently fail?  I might take a look at this eventually since my workaround was really really ugly.

mjambon:
@gtrak the technical issue is to avoid code duplication. Right now the `<json adapter.ocaml=...>` annotation is translated separately for each kind of node of the AST. In `oj_emit.ml`, we have:
```ocaml
      let adapter = j.json_sum_adapter in
      write_with_adapter adapter standard_writer
```
and for records we have:
```ocaml
      let adapter = j.json_record_adapter in
      write_with_adapter adapter standard_writer
```

Having `wrap` nodes in the core atd syntax is one way we solved this problem in the past, but I don't think they're suitable for use with json adapters. I also hope we don't have to create yet another kind of artificial node in the ast.

PR #153, opened 2018-09-06
Ocaml adapters support for bucklescript

rauanmayemir:
I started working on adapters support for bucklescript and made a little progress on decoding variants. (I haven't yet committed test cases before getting some initial feedback, but I played around with it on a real project and was able to decode fairly convoluted variant type)

Is there a reason to have separate `json_sum_adapter` and `json_record_adapter` fields? When are they used at the same time?

P.S: I tried to follow common indentation/formatting, but it looks arbitrary. Do you use `ocamlformat`?

### DIFF
--- a/atdgen-codec-runtime/src/decode.ml
+++ b/atdgen-codec-runtime/src/decode.ml
@@ -117,3 +117,6 @@ let option_as_constr f =
 let nullable f = function
   | `Null -> None
   | x -> Some (f x)
+
+let adapter (normalize: Json.t -> Json.t) (reader: 'a t) json =
+  reader (normalize json)
--- a/atdgen-codec-runtime/src/decode.mli
+++ b/atdgen-codec-runtime/src/decode.mli
@@ -38,3 +38,5 @@ val enum : (string * [`Single of 'a | `Decode of 'a t]) list -> 'a t
 val option_as_constr : 'a t -> 'a option t
 
 val nullable : 'a t -> 'a option t
+
+val adapter: (Json.t -> Json.t) -> ('a t) -> ('a t)
--- a/atdgen-codec-runtime/src/encode.ml
+++ b/atdgen-codec-runtime/src/encode.ml
@@ -86,3 +86,7 @@ let nullable f = function
 let option_as_constr f = function
   | None -> `Variant ("None", None)
   | Some s -> `Variant ("Some", Some (f s))
+
+let adapter (restore: Json.t -> Json.t) (writer: 'a t) x =
+  let encoded = writer x in
+  restore encoded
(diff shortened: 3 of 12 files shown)


mjambon:
> Is there a reason to have separate json_sum_adapter and json_record_adapter fields? When are they used at the same time?

Good question. The short answers are "it's an internal design issue that requires some extra thinking" and "they're not used at the same time; they're the same thing, duplicated".

Right now, each kind of node of the AST has its own set of options. For example, it makes sense for a variant type to support the option `<ocaml repr="classic">` which means "use classic variants instead of polymorphic variants". However, we don't have a good way to express that a given annotation applies to any kind of type. The special `wrap` construct kind of serves this purpose but it comes with its own meaning and I'm not sure it's the right solution here. So what I did is dodge the issue and decided that json adapters would be supported only for records and for variants, until we find the time to work on a better solution.

There are clearly use cases for json adapters outside of variants and records so this is on the roadmap. Obvious examples include simple types that don't follow json idioms (e.g. converting `"true"` to `true`) and unwrapped singletons (e.g. converting `"items": 42` to `"items": [42]`). The corresponding task is #130.

For the moment, I'd suggest you proceed with a similar technique and support json adapters only for records and for variants.

mjambon:
> P.S: I tried to follow common indentation/formatting, but it looks arbitrary. Do you use ocamlformat?

No, I don't like the results produced by ocamlformat. I use ocp-indent to normalize indentation, that's it. We can set up a pre-commit hook to run ocp-indent and check line length automatically, see https://github.com/mjambon/atd/issues/110

rauanmayemir:
Just added tests for adapters, @rgrinberg I think this is ready for review.

I'd like to limit the scope of this PR to sum type only, as I'm not sure how record adapters supposed to work.

rauanmayemir:
A nudge to remind about this PR. @mjambon @rgrinberg

rgrinberg:
Yeah, sorry about that. Let me give this an initial review right now.

rgrinberg:
I merged the PR with a couple of minor cleanups and analogous record adapter support. Ignoring the field entirely for records isn't right, and there's no need to error out if we can provide simple support.

One thing that's a bit worrying is that the interface of both runtimes is differing. That's an issue as we would like to them to converge to a single interface with multiple implementations at some point.

rgrinberg:
I merged the PR with a couple of minor cleanups and analogous record adapter support. Ignoring the field entirely for records isn't right, and there's no need to error out if we can provide simple support.

One thing that's a bit worrying is that the interface of both runtimes is differing. That's an issue as we would like to them to converge to a single interface with multiple implementations at some point.

PR #157, opened 2018-09-08
Fix #155

### LINKED ISSUE #155: atdgen -bs generates code with Atdgen_codec_runtime.Encode.tuple1 which doesn't exist
```
$ cat q.atd 
type t = [ A of (int) ]
$ atdgenbs -bs q.atd 
$ cat q_bs.ml
(* Auto-generated from "q.atd" *)
              [@@@ocaml.warning "-27-32-35-39"]

let write_t = (
  Atdgen_codec_runtime.Encode.make (fun (x : _) -> match x with
    | `A x ->
    Atdgen_codec_runtime.Encode.constr1 "A" (
      Atdgen_codec_runtime.Encode.tuple1
        (
          Atdgen_codec_runtime.Encode.int
        )
    ) x
  )
)
let read_t = (
  Atdgen_codec_runtime.Decode.enum
  [
      (
      "A"
      ,
        `Decode (
        Atdgen_codec_runtime.Decode.tuple1
          (
            Atdgen_codec_runtime.Decode.int
          )
        |> Atdgen_codec_runtime.Decode.map (fun x -> ((`A x) : _))
        )
      )
  ]
)
```
should be either added to runtime (as noop?) or not generated

rgrinberg:
@ygrek I think this should fix the issue

### DIFF
--- a/atdgen-codec-runtime/src/decode.ml
+++ b/atdgen-codec-runtime/src/decode.ml
@@ -71,6 +71,11 @@ let fieldDefault s default f =
     | None -> default
     | Some s -> s)
 
+let tuple1 a = function
+  | `Tuple [w]
+  | `List [w] -> a w
+  | _ -> raise DecoderError
+
 let tuple2 a b = function
   | `Tuple [w ; x]
   | `List [w ; x] -> (a w, b x)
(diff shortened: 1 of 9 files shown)


mjambon:
Yes, this looks correct.

Issue #152, opened 2018-09-03
Bucklescript backend fails with recursive records types


jorisgio:
Recursive records type like 
```ocaml
type container = {
  id: string;
  children: container list;
}
```
generate a decoder like

```ocaml

​let rec read__2 js = (
  Atdgen_codec_runtime.Decode.list (
    read_container
  )
) js
and read_container = (
  Atdgen_codec_runtime.Decode.make (fun json ->
    (
      ({
          id =
            Atdgen_codec_runtime.Decode.decode
            (
              Atdgen_codec_runtime.Decode.string
              |> Atdgen_codec_runtime.Decode.field "id"
            ) json;
          children =
            Atdgen_codec_runtime.Decode.decode
            (
              read__2
              |> Atdgen_codec_runtime.Decode.field "children"
            ) json;
      } : container)
    )
  )
)
```
but type of make is abstracted : 

https://github.com/mjambon/atd/blob/fa14d651cf4b760bc6c73350a5e9dcf96f4c2bdf/atdgen-codec-runtime/src/decode.mli#L3 

so even if the runtime uses a functional value as decoder type, compiler will reject `let rec`. This is working with normal yojson backend as a lambda is emitted.
Probably need to change the record decoder to emit explicit lambda : 
https://github.com/mjambon/atd/blob/fa14d651cf4b760bc6c73350a5e9dcf96f4c2bdf/atdgen/src/obuckle_emit.ml#L131

rgrinberg:
Yeah, a lambda will certainly be necessary. Would making the type of decoders/encoders concrete be an acceptable workaround for you? That would let us drop all the `make`'s and we'll have our lambdas. This isn't so great, but I'm not sure how to fix it otherwise as we'd need a way to check if an atd type is recursive and use `fix` appropriately.

@mjabmon do you have any advice here?

jorisgio:
I guess it is fine yes. We discussed this to see if it could be fixed un runtime but the change needs to happen in atd as the signature makes the type abstract

mjambon:
I don't understand why eta-expansion is not happening by just looking at the code in `obuckle_emit.ml`. `Ox_emit.is_function` should be renamed `Ox_emit.is_lambda` for clarity. Other than that I have to run the program and see what it's doing. I'll try to take a look later tonight.

mjambon:
https://github.com/mjambon/atd/pull/158 should trigger the eta expansion that wasn't happening here.

Generated code:
```ocaml
[...]

PR #158, opened 2018-09-09
Remove incorrect lambda annotation

mjambon:
This removes some incorrect lambda annotations in the tree of generated code. To reduce future confusion, I renamed the function `Ox_emit.is_function` to `Ox_emit.is_lambda`. The old name is still available but deprecated.

This should solve #152.

### DIFF
--- a/atdgen/src/ox_emit.mli
+++ b/atdgen/src/ox_emit.mli
@@ -29,7 +29,13 @@ val get_type_constraint
   -> ('a, 'b) Mapping.def
   -> string
 
+(** Determine whether the start of the given block of code was annotated
+    with the "fun" tag, indicating that it represents a lambda (anonymous
+    function). *)
+val is_lambda : Indent.t list -> bool
+
 val is_function : Indent.t list -> bool
+[@@deprecated "Please use 'is_lambda' instead of 'is_function'."]
 
 val needs_type_annot : _ expr -> bool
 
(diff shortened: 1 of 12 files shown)


mjambon:
I spoke too soon. I added a test, which doesn't pass. The fix above is valid, but we have a similar error on `Atdgen_codec_runtime.Encode.field write_...`.

atdgen/src/ox_emit.mli
```diff
@@ -29,7 +29,13 @@ val get_type_constraint
   -> ('a, 'b) Mapping.def
   -> string
 
+(** Determine whether the start of the given block of code was annotated
+    with the "fun" tag, indicating that it represents a lambda (anonymous
+    function). *)
+val is_lambda : Indent.t list -> bool
+
 val is_function : Indent.t list -> bool
+[@@deprecated "Please use 'is_lambda' instead of 'is_function'."]
```
rgrinberg: Can't we just delete this directly? I don't think this is part of the public API.

mjambon: sure.

mjambon:
Do you want to keep this? This was a good opportunity for me to see how things work but in the end I think we want to get rid of the `make` calls and make the type of encoders and decoders concrete.

mjambon:
... except for the renaming of `is_function` to `is_lambda`. I can make a simpler PR for that.

rgrinberg:
> ... except for the renaming of is_function to is_lambda. I can make a simpler PR for that.

This is certainly desirable.

PR #160, opened 2018-09-20
Fix hardcoded /bin/bash

aij:
`bash` is often not in `/bin`. Script wasn't using bash-specific features anyway.

If it did require Bash, `#!/usr/bin/env bash` would be more portable.


mjambon:
Good, thanks!

PR #162, opened 2018-09-26
WIP: Scala support

aij:
Very preliminary Scala support .

I cloned `atdj` and got it producing valid Scala code for some simple cases. (simple sums and records) It's defining case classes and a method to convert *to* JSON using [Argonaut](http://argonaut.io/).

TODO:

- [x] Support simple sums
- [x] Support simple records
- [x] Produce case classes
- [x] Output compiles :tada:
- [x] Conversion to JSON using [Argonaut](http://argonaut.io/)
- [x] Support complex records
- [x] Support complex sums
- [ ] Conversion from JSON
- [x] Test execution of output
- [x] Implement argonaut Codec instances (encoder only)
- [x] Support type aliases
- [x] Support tuples


plokhotnyuk:
@aij Ivan, why do you use Argonaut? 
There are more efficient alternatives for parsing JSON with Scala like jsoniter-scala or DSL-JSON for Java.

aij:
@plokhotnyuk Mainly just because it's what we're using at work, and I'm already familiar with it... At this stage, getting something working seems higher priority than making it fast. I certainly would like to leave space for future optimizations though. jsoniter-scala looks pretty cool.

I don't think it would be too hard to support multiple Json libraries in the future. (Either via flags at codegen time or implicits/"typeclasses" at runtime, though the later may not be possible with compiler  macros.)

Do you think it would  make more sense to rename `toJson` as `toArgonaut` and expose more generic APIs like `atdj`s `String toJson()` and `void toJsonBuffer(StringBuilder _out)`?

aij:
I'm not sure whether "forking" `atdj` is the best long-term maintenance strategy, and it certainly makes the diff here rather ugly. Here are the actual changes (excluding `cp`  and `sed`): https://github.com/mjambon/atd/compare/6472c8f7...aij:scala Let me know if there's a better way to fork a subdirectory in git. `git log --follow` at least understands `cp`, but `git blame` doesn't seem to.

atdj's TODOs are starting to make more sense to me, and I get the feeling it is somewhat lacking in features relative to atdgen. Is that why "merge into atdgen" is still on the list? I'm not yet sure how `Atd.Indent` would improve performance over `fprintf`, though it's pretty easy to imagine it improving clarity.

mjambon:
> atdj's TODOs are starting to make more sense to me, and I get the feeling it is somewhat lacking in features relative to atdgen.

Yes, a lot more time was spent on atdgen (ocaml/json) than atdj because it was used more extensively. I don't think all the features are necessary and some may be irrelevant in the java or scala worlds. From a maintainer's perspective, the "json adapter" feature that was released this year is a big relief. I'd consider providing something equivalent for scala (and other target languages) before working on more specialized features.

> Is that why "merge into atdgen" is still on the list?

There's a bit of duplication between atdj and atdgen but it's not necessarily a bad thing. I'd try to minimize the maintenance burden for our future selves, so try to reuse code if it helps in that direction.

> I'm not yet sure how Atd.Indent would improve performance over fprintf, though it's pretty easy to imagine it improving clarity.

The advantage of `Atd.Indent` is that you don't have to keep track of the amount of indentation when generating blocks of code. In atdgen, we also have `Annot` nodes which are used to perform minor rewrites of the generated code (in atdgen, specifically, we have to rewrite `let rec f = ... and g = ...` into `let rec f x = (...) x and g x = (...) x` depending on the nature of `...`, which is given by an `Annot` node).

aij:
@mjambon Is this good enough as a first pass? I'm tempted to work on Flow support next, and then come back to Scala.

Still missing features:

 * Tutorial
 * `wrap` types
 * Generics
 * JSON adapters
 * JSON decoding
 * Validators
 * Others?

mjambon:
I'm not in a good position to detect issues with the Scala side of things. Also, the OCaml code looks very readable, so I'm not worried about merging it. I'll post a few comments about little things that I noticed.

atds.opam
```diff
@@ -0,0 +1,19 @@
+opam-version: "2.0"
+maintainer: "martin@mjambon.com"
+authors: ["Martin Jambon"]
```
mjambon: You should put your name here.

atds/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+The primary benefits of using the generated interface, over using a 
+JSON library and writing case classes directly, are safety and ease of use
+when the same interface is needed in other languages.
+
+Don't repeat yourself when defining a JSON interface in multiple languages. 
+Instead generate compatible interfaces for all languages from a single definition.
+
+Even if you're only using Scala, you may find the ATD definitions more concise
+and less boilerplaity than the equivalent Scala types. (Especially for sum types)
```
mjambon: Feel free to throw in examples, information on the direction of the project, how to get started as a user, how to contribute, and other soft information. This can be done later, of course.

atds.opam
aij: Do you mean in the authors, or also maintainer?

I've actually never used opam before... The last time I was working mostly in OCaml, was back when oasis was just starting to catch on. 

I was assuming you would publish to opam along with the other packages when doing a release, but if you'd prefer I publish atds, I can look into how that works.

mjambon: I mean the authors.

Other people can take care of the rote aspects of publishing releases. What we'll need from you is that you update the root `CHANGES.md` with changes occurring in atds.

PR #168, opened 2018-10-29
Fix code block formatting

struktured:
This PR fixes a code block in `syntax.rst` that was accidentally using markdown formatting instead of rst formatting.

### DIFF
--- a/doc/syntax.rst
+++ b/doc/syntax.rst
@@ -495,15 +495,15 @@ any monomorphic type expression.
 It allows notably to represent cyclic values and to enforce that cycles
 are preserved during transformations such as serialization.
 
-```ocaml
-(* Example of a simple graph type *)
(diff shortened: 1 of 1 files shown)


mjambon:
thanks!

struktured:
@mjambon Above the block I fixed, it has this paragraph:

> ATD supports a special type $x$ shared where $x$ can be any monomorphic type expression. It allows notably to represent cyclic values and to enforce that cycles are preserved during transformations such as serialization.

What is the formatting intention of `$x$` here? It does not render it in rst in any special capacity.

mjambon:
The document was translated from LaTeX, where dollar signs enclose mathematical expressions. In this case it would just italicize the 'x'. I think the intent was to represent a placeholder for code rather than actual code, which why a normal monospace/code font was not used. I'll let you decide what's clearer in the context.

PR #173, opened 2018-11-20
Fix missing record field disambiguation type ascriptions

rdavison:
This is an attempt to fix https://github.com/mjambon/atd/issues/172

I'm not sure if this is the best way to fix the bug which I described in the issue, but if not, then perhaps it's a starting point.  As of right now, this branch correctly generates a compilable `swagger_j.ml` in https://github.com/rdavison/ocaml-swagger/tree/watch without having to apply the patch mentioned in the issue.


rdavison:
Superseded by https://github.com/mjambon/atd/pull/182

PR #178, opened 2018-12-02
Support jsoo compilation

Khady:
This PR replaces `str` with `re` so that atdgen binaries can be compiled using jsoo. It would offer an easy way to publish an atdgen executable to bucklescript users without depending on opam.


rgrinberg:
Looks straightforward. Thanks.

PR #188, opened 2019-03-15
Fix Yojson deprecation warnings

aij:
See https://github.com/ocaml-community/yojson/issues/79

    Error (warning 3): deprecated: Yojson.Safe.json
    json types are being renamed and will be removed in the next Yojson major version. Use type t instead

### DIFF
--- a/atdgen-runtime.opam
+++ b/atdgen-runtime.opam
@@ -16,5 +16,5 @@ depends: [
   "ocaml" {>= "4.02.3"}
   "dune" {build}
   "biniou" {>= "1.0.6"}
-  "yojson" {>= "1.2.1" }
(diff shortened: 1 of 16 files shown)


rgrinberg:
Thanks

Khady:
I think it deserves an entry in the changelog because it breaks compatibility with a big part of the opam world.

rgrinberg:
I've added a CHANGELOG entry here and in other places.

Khady:
Thinking about this more, I think that adding annotation to locally disable the deprecation warning would have been a better solution. atd would still be compatible with most of the yojson versions rather than the last one only. And this change is not future proof, as Safe is supposed to be remove in yojson 2. So there is no advantage to the use of `t` over `json`.

PR #192, opened 2019-03-21
Add Flow (JavaScript) support

aij:
`atdjs` generates Flow types for JavaScript. I expect it would be easy to adapt for TypeScript too.


rgrinberg:
Tbh, I'm not a fan of this way of adding new backends. In my experience, all this copy paste creates a serious maintenance overhead for future work. Would it be possible to properly factor the code between the flow and scala backends?

aij:
@rgrinberg TBH, I was a little dubious myself... I made the copy pasta commit separate to make the provenance of any copy pasted code very clear. (Same as with atdj -> atds.) In the end, there's less duplication than I expected. (Mainly [`Atdjs_main`](https://github.com/mjambon/atd/blame/6eb06150836f7b7465df2457241d6bfbabd7e1d2/atdjs/src/atdjs_main.ml) and [`Atdjs_names.to_camel_case`](https://github.com/mjambon/atd/blame/6eb06150836f7b7465df2457241d6bfbabd7e1d2/atdjs/src/atdjs_names.ml))

Scala and Java actually have more in common since they are both JVM languages, whereas JavaScript -- despite containing the substring "Java" in it's name -- is not.

I think it might actually make more sense to integrate this into `atdgen` than `atds`. I started from `atds` mainly because it is simpler and I was already more familiar with it, but I think `atdgen` has a better structure for supporting more advanced features like opaque types and json adapters.

My current plan for PureScript is to add it to `atdgen` directly, and then integrating `atdjs` when I end up wanting better feature parity. PureScript is probably more similar to Bucklescript than to anything else here. (It's basically a Haskell dialect in terms of syntax/static semantics, but with dynamic semantics based on JS.)

mjambon:
Hey @aij, what happened to your effort to add support for Flow? Did you end up using it?

aij:
We ended up not using it. Our architect decided we should roll our own instead, resulting in https://github.com/ccap/purescript-ccap-codegen

I've since switched jobs and currently have no plans to use Flow in the foreseeable future, so I should probably close this MR. I'm happy to advise if anyone wants to pick up where I left off.

mjambon:
Thanks for the update. I might start work on TypeScript support (atdts) soon, depending on the success of atdpy and other internal priorities at r2c.

PR #194, opened 2019-06-03
generate serializers for www-form data

cyberhuman:
This PR adds `-w` flag to generate serializers for www-form data. Serializer for type `t` is `www_form_of_t : t -> (string * string) list`. The resulting assoc list can be passed to e.g. `Netencoding.Url.mk_url_encoded_parameters` and sent either as a GET params or in a POST body with content type "application/x-www-form-urlencoded".
The output may be ambiguous in case of nested lists/arrays/records, since arrays are created by  appending `[]` to the field name. It could be solved in the future by having an annotation to number array elements (`[0]`, `[1]` etc).
Deserializers are not generated at this time.

p.s. I have also moved some boilerplate code to `Ox_emit`. I understand ideally the code must be refactored to reduce duplication between ocaml backends even more, although I'm not ready to do it at this time.


mjambon:
Hi, unfortunately I cannot afford to work on this project any longer. Hopefully someone with more time and enthusiasm will review your pull request. In the future, please let people know you'd like to add a new feature before putting much effort into it, especially if it involves a lot of code.

Regarding your work: You probably should explain why the feature you're adding is important, why it's better done as part of atd rather than a separate project, and who will be maintain that code.

cyberhuman:
It's very useful for various HTTP APIs that accept URL parameters or www-form-urlencoded body.
So instead of building that list of parameters manually, which is error-prone (string typing meh...), one could leverage the atdgen power.
Also, I was thinking about this kind of generator for quite some time already, but recently I had to work with Stripe API, and what surprised me is that they return JSON objects from their endpoints, but to create an object one has to supply it via form-encoded params. E.g. see https://stripe.com/docs/api/customers/create?lang=curl. With atdgen support, it's possible to use the same base type both to push and to pull data from Stripe.

mjambon:
Could we not have a single library function that converts a json object to
this, without any special-purpose code generation?

cyberhuman:
It is possible to convert a json object to www-form and this is how it was done initially, but 1) it requires unnecessary dependency on yojson and unnecessary serialization/deserialization pass, and more importantly, 2) we are going to implement generation of deserializers for www-form, and JSON workaround will not work anymore (e.g. it's not possible to distinguish string vs int when building a JSON from www-form data). 
If you're not willing to merge, it's ok. We will maintain a fork that has this PR merged.

rgrinberg:
I kind of agree with Martin that adding a separate code generator every single format is wasteful. Adding a dependency on yojson, does not seem to be a big deal to me. The main reason to want a dedicated code generator is to maximize the performance.

> we are going to implement generation of deserializers for www-form, and JSON workaround will not work anymore (e.g. it's not possible to distinguish string vs int when building a JSON from www-form data).

Could you explain this issue? Why do you need to distinguish integers and strings when deserializing? As far as I know, www-form only has strings so you might as well just use string.

With that being said, if the performance is important for your case, I think that it's a good enough reason to merge this additional generator.

I would mark it as experimental in the beginning however.

cyberhuman:
> Why do you need to distinguish integers and strings when deserializing?

Because, well, we want a well-typed data on the OCaml side. If the parameter is an integer, I want to provide an integer to the serializing code, and get back an integer from the deserializing code. That's the point of atdgen, isn't it? :-)

> I would mark it as experimental in the beginning however.

This is running in our production code for quite some time already and there is no issues.
I really miss deserializers though, but it's much harder to write them.

Khady:
We have a fork now living as an independent project https://github.com/ahrefs/atd-www

mjambon:
Closing based on @Khady's earlier message.

PR #198, opened 2019-12-03
Prepare 2.1.0 release

rgrinberg:
Use dune to generate all the opam files. This makes them a lot more
consistent.

@cyberhuman is the minimum version of 1.11 for dune ok? It makes it quite a bit easier to maintain the opam files.

If not, then  we can simply comment out the dune-project metadata just for the release.


cyberhuman:
Currently we are stuck on dune 1.10 because of some bugs that are only fixed in 2.0.
But I think it should not be a blocker for this release, since we are using a fork either way.

PR #208, opened 2020-02-24
tests must be restricted to a package

pirbo:
Else they are run while installing anyone of the packages of the repo which fails


pirbo:
Sorry, I'm playing with github web interface in order to see how far I can go without actually cloning the repo :-) The 2 commits have obviously to be splashed.

rgrinberg:
Looks good. Thanks

PR #210, opened 2020-05-06
Deny inline record type

### LINKED ISSUE #209: inline records create code with syntax errors
I was writing a ATD doc with something like the following:

`type foo = [ Foo of {x: int}]`

but this generates the incorrect OCaml code:

```ocaml
type foo = [`Foo of {x: int}]
```

I was expecting that ATD would give a syntax error on this.

dingxiangfei2009:
Fix #209 
It is one way to address this problem by denying anonymous record types at record field type or variant field type.

### DIFF
--- a/atd/src/parser.mly
+++ b/atd/src/parser.mly
@@ -90,19 +90,26 @@ type_var_list:
 ;
 
 type_expr:
-| OP_BRACK l = variant_list CL_BRACK a = annot
-     { Sum (($startpos, $endpos), l, a) }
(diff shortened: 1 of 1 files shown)


rgrinberg:
Thanks for your contribution. Could you add a test case for this?

rr0gi:
superseded by #224

Caprice2201 (commented):
46bd

Caprice2201:
disallow-inline-record-type

PR #213, opened 2020-07-08
Generate open! instead of open for external modules

ixzzd:
This PR  ~~disables 33 warning (unused open statement) for bucklescript emitter~~ generates `open!` instead of `open` for external modules.
This is helpful if you share atd generator command across atd files.
For example, I run atdgen for all `.atd` files in a folder, most of those files use types from some shared type module, but others are simple and don't use types from shared type modules, in this case, I want to silence unused open statement warning

### DIFF
--- a/atdgen/src/obuckle_emit.ml
+++ b/atdgen/src/obuckle_emit.ml
@@ -731,7 +731,7 @@ let make_ocaml_files
       | Some path -> sprintf "%S" (Filename.basename path)
     in
     sprintf {|(* Auto-generated from %s *)
-              [@@@ocaml.warning "-27-32-33-35-39"]|} src
+              [@@@ocaml.warning "-27-32-35-39"]|} src
   in
   let ml =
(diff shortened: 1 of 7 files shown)


ygrek:
I think the proper fix is to generate `open!` instead of `open` for external modules

ixzzd:
@ygrek I updated the PR, now it generates `open!` instead of `open` for external modules

rgrinberg:
Could you add a test that exercises this change?

ixzzd:
@rgrinberg I updated current tests, replaced `open` with `open!` where needed

ygrek:
It took actually using this code in real world to realize that `open!` still generates a warning (66) when it is unused (duh). i.e. my idea "to generate open! instead of open for external modules" was bad. Should generate annotation silencing warning (33 for `open` or 66 for `open!)

ixzzd:
I added one more PR, which revert current PR changes and silence 33 warning with annotation  https://github.com/ahrefs/atd/pull/214

Issue #219, opened 2021-07-08
Atdgen_runtime.Json_adapter.Type_field should  filter "type" field


dweil:
If I follow official documentation in https://atd.readthedocs.io/en/latest/atdgen.html#atd-annotations, the following code should be correct : 
```
type document = [
  | Image of image
  | Text of text
] <json adapter.ocaml="Atdgen_runtime.Json_adapter.Type_field">

type image = {
  url: string;
}

type text = {
  title: string;
  body: string;
}
```
Actually, I need to add a unseless fild "type : string" in both types "image" and "text".
Correction is trivial by just adding a filter in the adapter:
```
  let normalize (x : t) : t =
      match x with
      | `Assoc fields ->
          (match List.assoc type_field_name fields with
           | `String _type_ -> `List [ `String _type_; `Assoc (List.remove_assoc type_field_name fields) ]
           | exception Not_found -> x
           | _ -> x (* malformed *)
          )
      | `String _type_ as x -> x
      | malformed -> malformed
```

cyberhuman:
> Actually, I need to add a unseless fild "type : string" in both types "image" and "text".

Can you please demonstrate why you need to add it? We use atd and this particular adapter extensively and never ran into the issue. 
AFAIK this is a feature, not a bug. See https://github.com/ahrefs/atd/blob/9a3cb984a695563c04b41cdd7a1ce9454eb40e1c/atdgen-runtime/src/json_adapter.mli#L55-L56

dweil:
I use this adapter to translate a hierachy of JAVA classes (serialized in JSON) into OCAML variant types. For this, the json adapter described in the official documentation would be perfect. But with the current version of atd, I need to add the useless field "type : string" which will pollute my atd and my OCaml code The other solution is to write my own adapter .

But this implementation of this adapter is not compliant with the official atd documentation, this is why I logged the bug.

NB, if the issue and the pull-request, and if you really need this "type" field, you will have to write your own adapter. 

Below is the atd I had to write with the standard adapter
```
type document = [
  | Image of image
  | Text of text
] <json adapter.ocaml="Atdgen_runtime.Json_adapter.Type_field">

type image = {
  type : string;
  url: string;
}

type text = {
  type : string;
  title: string;
  body: string;
}
```

cyberhuman:
@dweil thank you for explaining the use case in detail! But I still miss the "why you need to add it" part; that is, what happens if you don't specify "type: string" in the `image` and `text` records in your example above?

[...]

PR #220, opened 2021-07-09
fix issue #219 : Atdgen_runtime.Json_adapter.Type_field should filter…

dweil:
… "type" field

### DIFF
--- a/atdgen-runtime/src/json_adapter.ml
+++ b/atdgen-runtime/src/json_adapter.ml
@@ -19,7 +19,7 @@ module Type_field = struct
       match x with
       | `Assoc fields ->
           (match List.assoc type_field_name fields with
-           | `String type_ -> `List [ `String type_; x ]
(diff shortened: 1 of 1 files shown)


atdgen-runtime/src/json_adapter.ml
```diff
@@ -19,7 +19,7 @@ module Type_field = struct
       match x with
       | `Assoc fields ->
           (match List.assoc type_field_name fields with
-           | `String type_ -> `List [ `String type_; x ]
+           | `String _type_ -> `List [ `String _type_; `Assoc (List.remove_assoc type_field_name fields) ]
```
rgrinberg, line 22: The leading underscore is reserved for intentionally unused identifiers. It shouldn't be used here.

rgrinberg:
Would you mind adding a test for this issue? See the existing tests for how to do this.

dweil:
Closing the pull-request as the issue has been closing as "won't fix", I'll write my own adapter.

PR #223, opened 2021-11-27
Specify (using menhir 2.0).

fpottier:
This causes Dune to pass --infer to Menhir.
This allows working around a type-checker bug in OCaml 4.07-4.10.


ygrek:
Thanks!

Issue #227, opened 2021-12-03
Add a way to check if annotation is valid and/or makes sense at a given position


zindel:
Inspired by https://github.com/ahrefs/atd/issues/204

For example:
```
type foo = {
  foo_bar: bool <json name="fooBar">;
}
```
It is obvious that user wanted to have the field named `fooBar` in the json output, but we have no way to report that right now. Another even more obvious case is:
```
type foo = {
  foo_bar<json nam="fooBar">: bool;
}
```
Again, `atdgen` is happy about it, while it shouldn't be ideally.

mjambon:
An approach I would consider is one that benefits all the target languages. We define a generic iterator/visitor over the ATD tree that visits all the annotations and call a user-specified function on the corresponding node + annotation. If I counted correctly, there are 6 types of nodes in the [AST](https://github.com/ahrefs/atd/blob/master/atd/src/ast.mli) that can bear annotations. So we'd define a visitor as something like this:

```ocaml
type 'acc annot_visitor = {
  type_def: type_def -> annot -> 'acc -> 'acc;
  type_expr: type_expr -> annot -> 'acc -> 'acc;
  ... (* 4 more fields or so *)
}

type node =
  | Type_def of type_def
  | Type_expr of type_expr
  | ...

val visit_annot : 'acc annot_visitor -> node -> 'acc -> 'acc
```
With this plus some helper functions, it should be relatively painless to define which annotations are valid and where, and call the visit_annot function to collect all the invalid annotations that were found.

Issue #204, opened 2019-12-25
Annotation in the wrong location should be an error


kittipatv:
I'm new to this library. It took me a while to debug the follow code, as it happily compiles

```
type foo = {
  foo_bar: bool <json name="fooBar">;
}
```

The annotation there has no effect. It would be nice to be warned about that.

```
$ atdgen -version
2.0.0
```

zindel:
That is an interesting one. In your example the annotation belongs to type `bool`, so there is no way now to tell what you wanted exactly. As a human being I totally understand your confusion - been there too.

 The bigger problem though is that we don't have an easy way to check if annotation makes sense at a given position. I'll add the separate issue for that. I think fixing it at a low level will fix your one too. The report could be something like `type annotation <json ...> is not applicable here`

PR #229, opened 2021-12-05
Validate annotations

oaons:
Trying to solve https://github.com/ahrefs/atd/issues/227 and https://github.com/ahrefs/atd/issues/204

- [x] added straightforward validation of `json` field names + tests. Now invalid field name will throw an error on `atdgen -j ...`
- [x] removed tests related to retired `json.tag_field` field since now they generate errors


rr0gi:
I have a gut feeling it should be a warning, not a hard error. E.g. : atd can have annotations for different languages, atdgen may not know all annotations interpreted by other generators.

atdgen/src/json.ml
```diff
@@ -63,6 +63,26 @@ type json_repr =
...
+  "name";
+  "keep_nulls"
+]
+
+let validate_json_annotation an =
+  List.iter (fun (s, (_, fs)) ->
+    if s = "json" then
+      List.iter (fun (f, (loc, _)) ->
+        if (not (List.mem f json_fields)) then
+          Error.error loc (Printf.sprintf "json.%s annotation not allowed" f)
```
rr0gi, line 82: s/not allowed/is not recognized/

rr0gi:
though this argument doesn't apply to json annotations.. Another weak argument - this removed tag_field annotation, we may want to make it possible process old atds without causing build breakage (it will still break later if code expects tag_field to work)

mjambon:
Hey all, thanks for starting this. I implemented schema-based validation for easily specifying which annotations of a given kind are allowed where. The PR is here: #244

cyberhuman:
Superseded by #244.

PR #231, opened 2022-01-07
Encode int64 value as string in json codec

zindel:
Currently, `atdgen-runtime` encodes `int64` value as JSON `number`. This makes it difficult to use the resulting JSON in the JavaScript environments (node, browser). For example, consider this issue with the `Int64.max_int` value:

```
Welcome to Node.js v12.16.2.
Type ".help" for more information.
> JSON.parse("9223372036854775807")
9223372036854776000
```

Right away, some information is lost. If we encode it as string though, we don't have such problem anymore. Moreover decoding already supports string as the transportation format for `int64` (test is added). This change fixes the interoperability with bucklescript https://github.com/ahrefs/bs-atdgen-codec-runtime/issues/22

Nevertheless, the change is breaking, so version bump is needed if merged.

### DIFF
--- a/atdgen-runtime/src/oj_run.ml
+++ b/atdgen-runtime/src/oj_run.ml
@@ -105,7 +105,9 @@ let write_int32 ob x =
   Bi_outbuf.add_string ob (Int32.to_string x)
 
 let write_int64 ob x =
-  Bi_outbuf.add_string ob (Int64.to_string x)
+  Bi_outbuf.add_char ob '"';
+  Bi_outbuf.add_string ob (Int64.to_string x);
+  Bi_outbuf.add_char ob '"'
 
 let min_float = float min_int
 let max_float = float max_int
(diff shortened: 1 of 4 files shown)


rr0gi:
change is breaking in the sense that on-the-wire json is changed?

zindel:
@rr0gi yes. Before `123`, after `"123"`

rgrinberg (approved):

Khady:
this needs a mention in CHANGES.md

AndreyErmilov:
@Khady @zindel  You forgot to mention this change in CHANGES.md :)

mjambon:
Sorry, I made the release (2.3.0) without trying hard to update the changelog. I think it's still a good idea to update it after the fact. You should be able to update it now in git, and edit manually the release page https://github.com/ahrefs/atd/releases/tag/2.3.0 (copy-paste the 2.3.0 section from `CHANGES.md` to replace what's already there).

Khady:
Sorry about that. I've updated the file and the release

Issue #184, opened 2019-02-06
Allow arguments to adapters


yonilevy:
[Feature Request]
[Bear with me, as I'm new to yojson/atd and OCaml in general]

I find adapters to be critical when working with external json structures, which I presume is a very common use case. The ability to pass arguments to adapters would make them much more flexible, readable within the .atd context, and allow for reuse and sharing of adapters.

Two use-case examples:
1) The provided `Atdgen_runtime.Json_adapter.Type_field` to take "type" field key as an argument
2) I've written an adapter to "lift" some fields, so `{"a": 1, "b": 2, "c": 3}` when lifting with `"body", ["a"]` would become `{"a": 1, "body": {"b": 2, "c": 3}}` -- it'd be great if i could pass those arguments in the .atd file.

Does that sound reasonable at all?

mjambon:
It seems reasonable to me. Keep in mind that json adapters are a relatively recent feature.

There are a few things to take into account:
* We could support annotations like `<json adapter.ocaml="Lift" adapter.ocaml_param="(\"body\", [\"a\"])">`. In this case, the presence of `ocaml_param` would indicate that the functions `normalize` and `restore` take its value as an extra argument. Another option I can think of would be to support `<json adapter.ocaml_normalize="Lift.normalize \"body\" [\"a\"]" adapter.ocaml_restore="Lift.restore \"body\" [\"a\"]">`.
* Having to escape double quotes isn't very nice but I don't know if there's a better solution without making things too complicated.
* We're short in personnel. If you or someone else could post first a detailed proposal and later the (simple) implementation with documentation, it would be ideal.

yonilevy:
Great!

Your first suggested syntax is what I had in mind, I agree the escaping problem is annoying but that it'd be best to start with that and perhaps improve in the future. I'll try to dive in and see if I can come up with something.

mjambon:
Implemented by https://github.com/ahrefs/atd/pull/232

PR #232, opened 2022-01-07
Support arbitrary OCaml code in adapter.ocaml

zindel:
**Update 06.02.2022**: This one can be reviewed. The only missing piece is the documentation entry in `doc/atdgen.rst`, but I'd rather add it if the functionality itself is accepted.

This PR is solves the https://github.com/ahrefs/atd/issues/184 exactly as @mjambon proposes [here](https://github.com/ahrefs/atd/issues/184#issuecomment-461137728) (second option).

Here is an example inspired by the original issue description, assuming we have two functions:
-  `Lift.normalize rest keys json` which transforms the provided `json` object  leaving only keys listed in `keys` in it, moving other keys to `rest`. (`List.normalize "body" ["a"] {"a":1, "b":2, "c":3}` -> `{"a":1, "body":{"b": 2, "c":3}}`)
- `Lift.restore rest keys json` doing the vice versa transformation

Example of atd:
```
type body = {  b: int;
  c: int;
}
type adapted = {
  a: int; 
  body: body;
}<json  adapter.ocaml_normalize="Lift.normalize \"body\" [\"a\"]"
  adapter.ocaml_restore="Lift.restore \"body\" [\"a\"]">
```

In `_j.ml` it looks like:
```OCaml
...
let write_adapted : _ -> adapted -> _ = (
  Atdgen_runtime.Oj_run.write_with_adapter (Lift.restore "body" ["a"]) (
    fun ob (x : adapted) ->
...
let read_adapted = (
  Atdgen_runtime.Oj_run.read_with_adapter (Lift.normalize "body" ["a"]) (
    fun p lb ->
...
```

Few notes:
- backward compatibility is preserved (`"adapter.ocaml"` keeps working as before)
- old & new way of adapter specification are mutually exclusive and error is triggered in case of unexpected fields combination (e.g. valid spec is either single `"adapter.ocaml"` or the pair `"adapter.ocaml_normalize"`+`"adapter.ocaml_restore"`, all the other combinations produce an error)

~~Unless there are any concerns with the approach I'll proceed with the documentation and tests~~ Done.


rr0gi:
<somewhat bikeschedding>
I am always confused about data flow in such annotation, maybe rename to from_ocaml and to_ocaml?

mjambon:
@zindel It looks good to me. What do you think of @rr0gi 's suggestion of supporting the names `from_ocaml` and `to_ocaml`? This could be in addition to `ocaml_normalize` and `ocaml_restore`:
```ocaml
<json adapter.ocaml_normalize="Lift.normalize \"body\" [\"a\"]"
  adapter.ocaml_restore="Lift.restore \"body\" [\"a\"]">
```
=
```ocaml
<json adapter.to_ocaml="Lift.normalize \"body\" [\"a\"]"
  adapter.from_ocaml="Lift.restore \"body\" [\"a\"]">
```

The original adapter mechanism requiring a module providing `normalize` and `restore` would be unchanged. Also, it might not get used very much since the new way is more flexible.

zindel:
I like the the explicitness of the `adapter.{from_ocaml/to_ocaml}` naming indeed! And the fact that it scales onto different languages as well.  Should we support `adapter.{ocaml_normalize/ocaml_restore}` though? What bothers me is that one could specify any subset of the 5 attrs (`adapter.ocaml` + `to_/from_` + `normalize/restore`) which we should be validating properly.

Maybe  we can simply rename the attrs (and leave the currently working module-based mechanism untouched)?

mjambon:
> Maybe we can simply rename the attrs (and leave the currently working module-based mechanism untouched)?

sounds good i.e. only add support for `adapter.{from_ocaml/to_ocaml}`, not for `adapter.{ocaml_normalize/ocaml_restore}`.

zindel:
Renamed attributes just as @mjambon & @rr0gi suggested above. Added docs/changelog entry. Please, check. If everything is ok - will merge

CHANGES.md
```diff
@@ -1,3 +1,9 @@
+2.6.1 (2022-XX-XX)
+------------------
```
mjambon, line 2: Small thing: we're adding a new feature so we should increase the minor version i.e.  `2.6.x` -> `2.7.0`. Note that I usually just write "Next release" instead of trying to guess the next version.

mjambon (approved):
Great. Thank you!

PR #234, opened 2022-01-10
Add C++ Atdgen

caiowakamatsuahrefs:
This adds basic C++ Atdgen, there are some types missing 
Most notably
```ocaml
type a = [ A of b list ]
type b = [ B ]
```

What is implemented currently is enough to have basic C++ generation.


atdcpp/src/atdcpp_names.ml
```diff
@@ -0,0 +1,155 @@
...
+  "yield";
+]
+
+let is_cpp_keyword =
+  let tbl = Hashtbl.create 200 in
+  List.iter (fun k -> Hashtbl.add tbl k ()) cpp_keywords;
+  fun k -> Hashtbl.mem tbl k
+
+(*
+   Automatically append an underscore to a field name if it is a Scala keyword.
```
Khady, line 88:
```suggestion
   Automatically append an underscore to a field name if it is a C++ keyword.
```

caiowakamatsuahrefs: resolved, also updated list of keywords for C++

atdcpp/src/atdcpp_main.ml
```diff
@@ -0,0 +1,74 @@
...
+open Atdcpp_env
+
+let args_spec env = Arg.align
+    [ "-package",
+      Arg.String (fun x -> env := { !env with package = x }),
+      " Package name of generated files";
+      "-o",
+      Arg.String (fun x -> env := { !env with output = open_out x }),
+      " File name for C++ output"
+    ]
```
mjambon, line 13: I recommend using cmdliner, like I did in atdpy for a better user interface. Copy-paste-edit is usually the way to go.

mjambon:
Regarding name management, for atdpy I created a [`Unique_name`](https://github.com/ahrefs/atd/blob/d78f11508f61d1fda60f8140faee18074f674e89/atdpy/src/lib/Unique_name.mli) module which I was thinking of moving to the `atd` library since it's generic and was properly tested and documented. Not sure if this would work well for C++ but maybe it does. Or if it doesn't, I'd love to know what's missing.

PR #238, opened 2022-02-06
Allow to specify type attributes in the atd file

zindel:
Currently `atdgen` permits to specify ppx_deriving extentions or in general any attributes using command line arguments: `-type-attr`, `-type-conv`, `-deriving-conv`. The problem is that it might be an overkill for some use-cases. This experimental PR adds an ability to specify those on per-type basis right in the `.atd` file. Currently, the per-type attributes overload those defined globally, this could be changed of course.

Example:
```
$ cat x.atd && _build/default/atdgen/bin/ag_main.exe -t x.atd && cat x_t.ml
type x = {
  v: int;
}<ocaml attr="deriving show,eq" attr="ocamlformat \"disable\"">


(* Auto-generated from "x.atd" *)
              [@@@ocaml.warning "-27-32-33-35-39"]

type x = { v: int } [@@deriving show,eq][@@ocamlformat "disable"]
```

Please, let me know if the direction makes sense with the regard to consistency of the overall design.


rr0gi:
one day somebody will want to add `@`-attributes, so maybe it makes sense to be more explicit (and also uniform with cmdline option) - `ocaml.type-attr`

mjambon:
I'm not super familiar with attributes, so please correct me if I got things wrong. My understanding is that `[@@foo]` attaches to a whole type definition (`type ... = ...` or `and ... = ...`) whereas `[@bar]` attaches to a type expression. ATD supports annotations on a type definition (`type t <foo> = ...`) or on an a type expression (`type type t = int <bar>`). The third kind `[@@@baz]` is a top-level construct that doesn't attach to anything.

So, it seems that this example wouldn't work because it would produce an annotation `@` on the type expression ` { ... }` rather than on the whole definition:
```
type x = {
  v: int;
} <ocaml attr="deriving show,eq" attr="ocamlformat \"disable\"">
```
If I understand correctly, the above should instead be written like this:
```
type x <ocaml attr="deriving show,eq" attr="ocamlformat \"disable\""> = {
  v: int;
}
```

@rr0gi , I don't think we have a conflict between `@` and `@@` attributes because they would be attached to different nodes in the ATD tree. That said, we could still require the `@` and `@@` in the ATD annotation to make things clear. I don't have a strong opinion.

zindel:
@mjambon Thanks for the clarification on the type definition vs. type expression annotations. Although, I've tested & it worked in the resulting code, no matter where do we put them in atd. But your comment makes the proper distinction, so I followed it and now only consider the annotation from the type definition. Here is an example:
```
$ cat x.atd && _build/default/atdgen/bin/ag_main.exe -t x.atd && cat x_t.ml
type x  = {
  v: v;
}

type v <ocaml attr="deriving show,eq" attr="ocamlformat \"disable\""> = [ Node | Branch of x ] <ocaml repr="classic">
(* Auto-generated from "x.atd" *)
              [@@@ocaml.warning "-27-32-33-35-39"]

type v =  Node | Branch of x  [@@deriving show,eq][@@ocamlformat "disable"]

and x = { v: v }
```
The `X_t` module contains all the expected functions : `show_v, pp_v, equal_v, show_x, pp_x, equal_x`. 

Regarding the `@` vs `@@` distinction - I actually tried to keep consistency with the `-deriving-conv` & `-type-attr` arguments. Atm, those don't expect the leading @@ and I decided to follow the current convention. Can be reconsidered if anyone has strong opinion on it. I'm leaning towards skipping them for now. Can even add a validation step to ensure user won't do this by accident.

@rr0gi I really like your `type-attr` suggestion, but the implementation is more involved, since it requires the parser/lexer modifications. Currently, attribute name is the `LIDENT` token (identifier with the first lowercase letter or underscore), which follows the OCaml identifier convention and used in other places of parser as well, so `-` char is impossible there. Not saying, this is non-doable, but I'd rather do it if we have an agreement that we like the direction.

mjambon:
> Although, I've tested & it worked in the resulting code, no matter where do we put them in atd.

Ah yes, this is an interesting consequence of the OCaml annotations for the whole definition (`@@`) or for the type expression (`@`) being accepted at the same location. I predict the following ATD definition would produce invalid OCaml syntax:

```
type foo = int <ocaml attr="deriving show,eq"> list
```
because the following OCaml code is invalid:
```ocaml
type foo = int [@@deriving show,eq] list
```

mjambon (approved):

mjambon:
Don't forget to add some documentation to `/doc/`.

mjambon:
I pushed updates to your branch:
* merged master
* added entry to the new schema for validating annotations
* added test

mjambon:
I also added an entry in the reference manual and in the changelog. I'm merging this as soon as the tests pass.

PR #244, opened 2022-02-11
Add system for checking the placement of annotations

### LINKED ISSUE #204: Annotation in the wrong location should be an error

### LINKED ISSUE #227: Add a way to check if annotation is valid and/or makes sense at a given position

mjambon:
This catches misplaced annotations (typically those applying to field names and variant names) for the selected target(s). This validation is done for atdgen (ocaml, bucklescript, json, biniou) and atdpy (python/json). This caught a bug in an existing ATD test file!

Closes #204
Closes #227  

### PR checklist


atdgen/test/spec_js/spec.atd
```diff
@@ -22,7 +22,7 @@ type v1 =
   [ V1
   | V2 <json name="v22">
   | V3 of string
-  | V4 of int <json name="v44">
+  | V4 <json name="v44"> of int
   | V5 of bool option]
```
mjambon, line 26: This is the bug that was caught with the validator.

mjambon:
Thoughts anyone?

The only issue I can think of is that some existing builds will fail if some atd files contain invalid, useless annotations. The solution in that case is to remove the useless annotations. I don't think there's much value in adding a command-line option to disable validation.

cyberhuman (approved):

cyberhuman:
Nice!

> The only issue I can think of is that some existing builds will fail if some atd files contain invalid, useless annotations. The solution in that case is to remove the useless annotations. I don't think there's much value in adding a command-line option to disable validation.

It sounds good. Thinking about a possible solution, the error might optionally be a warning instead. But it seems not very useful.

cyberhuman:
@mjambon thanks for implementing this!

Issue #241, opened 2022-02-10
Set up continuous testing (CI)


mjambon:
It would be good to run all the tests automatically on each pull request. This will at least force us to have a working setup script that installs everything that's needed to test all the targets (ocaml, java, scala, python, ...).

PR #247, opened 2022-02-18
Add support for OCaml 5.00

kit-ty-kate:


cyberhuman (approved):
@kit-ty-kate it looks good, thanks! It would be nice to have a short summary though:

- why was the minimum required version of OCaml raised from 4.02 to 4.03
- why was the dependency on camlp-streams added.

kit-ty-kate:
`String.*case_ascii` is only available since 4.03 and this package makes use of the `Stream` or `Genlex` modules that got removed from the standard library in 5.00 (camlp-streams is the compatibility package)

mjambon:
Just a heads-up: I might have inadvertently raised the minimum ocaml version requirement two weeks ago when I merged the atdpy branch (e.g. use of `Fun.protect` which requires 4.08). We need to fix it before attempting and failing to release. A preliminary step would be to have CI for the atd project (#241).

kit-ty-kate:
Raising everything to 4.08 would also be a totally fine outcome IMO. Less problems with `Stdlib` and many more things ^^

mjambon:
I'm preparing for release 2.3.0. Minimum ocaml version will be 4.08, I think.

PR #248, opened 2022-02-24
Docker build + CircleCI config + cmdliner fix

mjambon:
This adds a way to build and test atd with each pull request.

* `make docker` will build and test atd locally in a Docker container. I split this into two parts because the first part which installs the ubuntu and opam packages takes a while (several minutes) while the second part only builds and tests atd and is relatively fast. This Docker setup allows local testing independently of whichever CI service we use.
* a CircleCI config which I [tested on returntocorp/atd](https://app.circleci.com/pipelines/github/returntocorp/atd). It uses the same base image as the Docker build and runs the same scripts. It tests against ocaml 4.08 and 4.13.

Along the way, I fixed the incompatibilities with the latest cmdliner (1.1.x) which was released two weeks ago.

Action needed:
- [ ] Merge this PR so as to put the CircleCI config on the `master` branch.
- [ ] Then, someone from the Ahrefs organization needs to enable CircleCI on the `atd` repo. It should be straightforward since the config is ready to use. CircleCI will find it on the `master` branch.

PS: I'm not sure who's both at Ahrefs and involved with atd. I picked reviewers who seem to be with Ahrefs (except for @rgrinberg).

### PR checklist


Khady:
For ahrefs, you can in general ask @zindel @cyberhuman @ygrek or me

Just curious, is there a particular reason to use circle ci instead of github actions?

.circleci/setup-opam
```diff
@@ -0,0 +1,16 @@
...
+# This is used by the Dockerfile and by the CI build.
+#
+set -eu
+
+eval "$(opam env)"
+
+opam repo add --all-switches -k git \
+  github https://github.com/ocaml/opam-repository.git
+opam repo remove default
+# no need for 'opam update'
```
Khady, line 14: is it faster than doing just `opam update` or `opam update -a`? Seems a bit overkill to clone the whole repo every time

.circleci/config.yml
```diff
@@ -0,0 +1,49 @@
...
+          command: ./.circleci/setup-system
+      - run:
+          name: Install opam dependencies
+          command: ./.circleci/setup-opam
+      - run:
+          name: Build
+          command: opam exec -- make
+      - run:
+          name: Test
+          command: opam exec -- make test
```
Khady, line 42: I think that you can remove the duplication. Circle ci supports matrix (probably the nicest solution) and commands (easiest one). I don't know the exact syntax though

An example with commands
```suggestion
commands:
  build:
    steps:
      - checkout
      - run:
          name: Install system dependencies
          command: ./.circleci/setup-system
      - run:
          name: Install opam dependencies
          command: ./.circleci/setup-opam
      - run:
          name: Build
          command: opam exec -- make
      - run:
          name: Test
          command: opam exec -- make test

jobs:
  build_latest_ocaml:
    docker:
      - image: ocaml/opam:ubuntu-20.04-ocaml-4.13
    working_directory: ~/atd
    steps:
      - build:

  build_min_ocaml:
    docker:
      - image: ocaml/opam:ubuntu-20.04-ocaml-4.08
    working_directory: ~/atd
    steps:
      - build:
```

.circleci/setup-opam
mjambon: The Docker image comes configured only with a local repository of opam packages (`~/opam-repository`) rather than the normal opam-repository on GitHub. I'm not sure why.

```
$ docker run -it ocaml/opam:ubuntu-20.04-ocaml-4.13
opam@6a6e0c37bd18:~$ ls
opam-repository
opam@6a6e0c37bd18:~$ date       
Thu Feb 24 06:08:30 UTC 2022
opam@6a6e0c37bd18:~$ ls -l
total 4
drwxr-xr-x 5 opam opam 4096 Feb 19 23:17 opam-repository
```
shows that the local repo is 5 days old, which is great, but I don't know if it's just luck or if it gets refreshed often enough. https://hub.docker.com/r/ocaml/opam is the home of those Docker images but I don't see the info about the freshness we can expect from the pre-downloaded opam-repository.

mjambon:
> Just curious, is there a particular reason to use circle ci instead of github actions?

I find CircleCI easier to work with. In particular, I like the documentation and the fact that we can ssh into a running instance.

Also, GitHub Actions [messes things up in our own Docker container and we need to fix it](https://github.com/returntocorp/semgrep/blob/938aac44b00d0bdc0f428e11209d48f7354766f7/.github/workflows/build-tests.yml#L20-L25). This gave me trouble in the past. I'm sure I don't understand something that started with a great sentiment... but that's the problem. I don't have difficulties understanding CircleCI.

For this project, I don't mind if we use one or the other, or something else. My preference is to keep the CI-specific config/code as simple as possible so as to (1) minimize CI vendor lock-in and (2) keep the ability to run the same flow locally to make debugging easier (via Docker).

.circleci/config.yml
mjambon: nice. I see the matrix feature is [demonstrated here](https://circleci.com/blog/circleci-matrix-jobs/), but let's use your solution.

Khady:
It looks good to me, thanks. I don't have the rights to enable circle ci, I've asked internally.

mjambon:
Thank @Khady. Merging!

Khady:
For the record it was enabled and passed https://app.circleci.com/pipelines/github/ahrefs/atd/1/workflows/771b03a3-994c-4760-bb21-3b23179b3e18/jobs/2

mjambon:
That's great. This PR adds a cute badge: https://github.com/ahrefs/atd/pull/250

PR #252, opened 2022-03-10
atdpy: avoid errors from the flake8 linter

mjambon:
A special comment is now inserted in the generated python files to disable the flake8 linter. We now run flake8 on our test files to check that everything's fine. It needs to be installed like other tools pytest and mypy.

### PR checklist


mjambon:
I'll be merging my own PR when I think they're uncontroversial. Let me know if that's a problem.

PR #255, opened 2022-03-11
release 2.3.0

mjambon:
I think this is ready for release. I still don't understand how topkg is supposed to work. `topkg distrib` fails, telling me it wants a `pkg/pkg.ml`. Also, there's no git tag for the previous release 2.2.1 or no trace of in `CHANGES.md` which is weird. I don't know if you remember what you did @rgrinberg for that release, which was [two years ago](https://github.com/ahrefs/atd/releases/tag/2.2.1).

I remember that one of the motivations for using topkg was that the github-generated tarball might change, which is bad for package managers that expect a particular checksum for the tarball. So I'll try to find a way to upload a tarball as an asset and use that URL in opam-repository.

### PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 Next release
 ------------
 
+* ...
+
+2.3.0 (2022-03-10)
+------------------
+
 * Allow single-quoted strings as an alternative to double-quoted
   strings in ATD files (#239)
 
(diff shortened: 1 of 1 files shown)


Khady:
We should probably use dune-release instead of topkg. It also takes care of creating the tarball and uploading it on github.

mjambon:
Taking a look at dune-release now.

mjambon:
dune-release worked wonderfully (the only glitches I saw are the tarball name and opam-repository PR title, which are named after one of the opam packages rather than after the git project).

The opam-repository PR is here: https://github.com/ocaml/opam-repository/pull/20898 (I botched the first one, 2.3.0)

PR #262, opened 2022-03-23
atdpy: add support for extra imports and class decorators

mjambon:
This will allow users to "enrich" the generated Python classes using decorators.

### PR checklist

### DIFF
--- a/doc/atdpy.rst
+++ b/doc/atdpy.rst
@@ -342,3 +342,44 @@ Python dictionaries if the correct annotations are provided:
 * ``(foo * bar) list <python repr="dict">`` will use a Python
   dictionary of type ``Dict[Foo, Bar]`` to represent the association list.
   Using the annotation ``<python repr="list">`` is equivalent to the default.
+
+
+Additional imports
+^^^^^^^^^^^^^^^^^^
+
+At the beginning of the ATD file, placing annotations like this one
+allow inserting arbitrary Python code or comments:
+
+::
+
+   <python text="import deco">
+
+This is the recommended mechanism for inserting imports. In contrast, it
+should be used only as last resort for inserting functions or classes.
+
+
+Custom class decorators
+^^^^^^^^^^^^^^^^^^^^^^^
+
+Extra class decorators can be specified in addition to ``@dataclass``.
+The following ATD definition will add 3 decorators:
+
+.. code-block:: ocaml
+
+   type thing <python decorator="deco.deco1"
+                      decorator="deco.deco2(42)"
+                      decorator="dataclass(order=True)"> = {
+     foo: int;
+     bar: string;
+   }
+
+The generated Python class will start like this:
+
+.. code-block:: python
+
+   @deco.deco1
+   @deco.deco2(42)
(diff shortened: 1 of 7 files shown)


doc/atdpy.rst
```diff
@@ -342,3 +342,44 @@ Python dictionaries if the correct annotations are provided:
...
+allow inserting arbitrary Python code or comments:
+
+::
+
+   <python text="import deco">
+
+This is the recommended mechanism for inserting imports. In contrast, it
+should be used only as last resort for inserting functions or classes.
+
+
```
Khady, line 360: looks pretty similar to the `-open` used for ocaml. We could probably add `<ocaml text...>` one day too

Khady (commented):
I don't know python very well so no real opinion on the decorators. Code looks ok

doc/atdpy.rst
mjambon: Ah, thanks for pointing that out. I actually thought we already supporte this. `<ocaml text="(* hello *)">` is indeed not rejected but also it's not inserted into the generated files.

So here's my mistake:
* we support `<doc text="whatever">`, not `<ocaml text="whatever">`
* the validator is fine with `<ocaml text="whatever">` because I mistook `doc` for `ocaml` when I created the validator's schema (atdgen 2.3.x)
* yesterday, I used the [validator's schema](https://github.com/ahrefs/atd/blob/4f574073d1902866bf7e4c03a213be6925feb29c/atdgen/src/ocaml.ml#L145) as a reference but it was wrong
* I replicated the feature for Python in atdpy without rethinking it

I didn't think too hard about this because I thought we were already using such feature with atdgen and that it would work just fine.

Now I think a problem with this feature is when we generate more than one file but the code needs to be inserted in one file and not another (e.g. `.ml` but not `.mli`). With atdpy, we produce only a `.py` file and we only deal with JSON, so that's fine for now. It will be problematic when we start generating different kinds of file for Python, like we do for OCaml.

I see the following possibilities:
1. accept that `<python text="abc">` will insert `abc` into all the generated Python files, which may be a problem in the future if we ever generate different kinds of files. This is however simpler for the user of atdpy.
2. offer finer-grainer options specifying the target files e.g. `<python json_py.text="abc">`
3. [require that `<python json_py.text="abc">` be specified rather than just `<python text="abc">`]

I think (1) and (2) are nice. I wouldn't want to be forced to write `<python json_py.text="abc">` unless necessary. For the OCaml backend, it's a similar story but we don't have much of a choice unless the text to insert is a generic comment that should go into all the files. Code would be inserted only into the `_j.ml` file with `<ocaml json_ml="abc">`.

Here's what I'm going to do:
* keep the behavior of `<python text="abc">`
* update the documentation to say it's inserted into all the generated files
* add support for the more correct `<python json_py.text="abc">`
* create a GitHub issue for doing the same thing for atdgen/OCaml (where we generate at least 8 kinds of files: ml/mli * _t/_j/_b/_v)

mjambon: ok, I'm done with the tasks above.

PR #267, opened 2022-03-25
atdpy fix: Avoid applying the `@dataclass` decorator twice

mjambon:
The generated Python class always applies the `@dataclass` decorator. However, the user may want to apply it with special options e.g. `@dataclass(frozen=True)`, which they do via an ATD annotation. This previously resulted in
```
@dataclass(frozen=True)
@dataclass
class Foo:
    ...
```
I previously thought that it was fine but it doesn't work, resulting in a runtime error indicating that the frozenness was violated. This PR ensures that `@dataclass` is only added if not explicitly added by the user.

### PR checklist

### DIFF
--- a/atdpy/src/lib/Codegen.ml
+++ b/atdpy/src/lib/Codegen.ml
@@ -1120,7 +1120,7 @@ let sum env ~class_decorators loc name cases =
   |> double_spaced
 
 let uses_dataclass_decorator =
-  let rex = Re.Pcre.regexp "\\A[ \t\r\n]*dataclass(\\(|[ \t\r\n]|\\z)" in
+  let rex = Re.Pcre.regexp {|\A[ \t\r\n]*dataclass(\(|[ \t\r\n]|\z)|} in
   fun s -> Re.Pcre.pmatch ~rex s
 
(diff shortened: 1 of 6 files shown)


atdpy/src/lib/Codegen.ml
```diff
@@ -1122,11 +1119,24 @@ let sum env ~class_decorators loc name cases =
   ]
   |> double_spaced
 
+let uses_dataclass_decorator =
+  let rex = Re.Pcre.regexp "\\A[ \t\r\n]*dataclass(\\(|[ \t\r\n]|\\z)" in
```
Khady, line 1123:
```suggestion
  let rex = Re.Pcre.regexp {|\A[ \t\r\n]*dataclass(\(|[ \t\r\n]|\z)|} in
```

just a style suggestion, when working with regexp I like to avoid the double escaping that is necessary with basic strings

Khady (approved):

atdpy/src/lib/Codegen.ml
mjambon: great. I don't have this reflex yet.

PR #268, opened 2022-03-25
Fix atd opam build template

Khady:
The script used to patch the opam files has a sed command to replace the old `pinned` syntax with `dev`. This part is actually not generated by dune but comes for a template file that can be directly fixed instead of relying on an external script.

### PR checklist


mjambon:
oh, nice. I had no idea.

PR #274, opened 2022-04-11
Add TypeScript support (atdts)

mjambon:
### PR checklist


mjambon:
Basic support for TypeScript was implemented and tested. I'll write the documentation in another PR. This PR is ready for merging.

.circleci/setup-system
```diff
@@ -13,11 +13,17 @@ sudo apt-get update
...
   default-jdk \
+  nodejs \
+  npm \
   python3 \
   python3-pip \
   python-is-python3 \
   scala
 
+# For atdts
+sudo npm install -g typescript
```
Khady, line 24: it's a bit suspicious to have it install globally. I think that it should be a dev depedency in a package.json and then have a `script` field in this same package.json that would be like

```
"scripts": {
  "tsc": "tsc"
}
```

atdts/src/lib/Codegen.ml
```diff
@@ -0,0 +1,1150 @@
...
+    | 'a'..'z' as c when !start_word ->
+        Buffer.add_char buf (Char.uppercase_ascii c);
+        start_word := false
+    | c ->
+        Buffer.add_char buf c;
+        start_word := false
+  done;
+  Buffer.contents buf
+
+(* Translate a preferred variable name into an available Python identifier. *)
```
Khady, line 62:
```suggestion
(* Translate a preferred variable name into an available TypeScript identifier. *)
```

atdts/src/lib/Codegen.mli
```diff
@@ -0,0 +1,7 @@
+(*
+   Python code generation for JSON support (no biniou support)
```
Khady, line 2:
```suggestion
   TypeScript code generation for JSON support (no biniou support)
```

atdts/src/lib/Codegen.mli
```diff
@@ -0,0 +1,7 @@
...
+*)
+
+(** Take ATD type definitions and translate them to Python, writing
+    them out to a file which should have the '.py' extension. *)
```
Khady, line 6:
```suggestion
(** Take ATD type definitions and translate them to TypeScript, writing
    them out to a file which should have the '.ts' extension. *)
```

Khady (approved):
You are using native typescript json decoding without relying on an external runtime as it was done for bucklescript?

There's a decent amount of duplicated code between tools now. But can be solved another time.

I didn't look for bugs or really checked the runtime

mjambon:
Sorry about all those mentions of python where it should be typescript. I should have fixed them before submitting the PR.

mjambon:
> You are using native typescript json decoding without relying on an external runtime as it was done for bucklescript?

The runtime is included in each generated `.ts` file. It's about [300 lines](https://github.com/ahrefs/atd/blob/52ecbd530f5626bf1dc8b8cf2f767bea6aecf702/atdts/test/ts-expected/everything.ts#L143-L475). We could have it in a runtime library but it's more complicated.

> There's a decent amount of duplicated code between tools now. But can be solved another time.

Yes

Khady:
> JSON parsing and printing is meant to be done by the [built-in JSON library](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON) (string <-> any). The code we generate validates and converts between the any value (JSON AST) and a typed value. A few things such as field names being renamed or the representation of variants change in the conversion, so it's not a pure validation. I'll try to explain this a bit in the documentation.

Sounds good. The goal in the bucklescript implementation was also to use the built-in JSON lib for good perf. Embedding the runtime makes it easier to use.

Khady:
CHANGES.md is ticked but is not updated in this PR

mjambon:
I changed the typescript/javascript setup to install dependencies locally + switched to strict type checking mode.

Merging!

.circleci/setup-system
mjambon: done.

PR #285, opened 2022-04-28
Add a translator from ATD to JSON Schema

### LINKED ISSUE #284: Translate ATD definitions to JSON Schema
For better or for worse, some people have asked for a converter from ATD to [JSON Schema](https://json-schema.org/).

Benefits include:
* Comparison between the two languages for a given interface
* Some support for languages not already supported by ATD

mjambon:
I think this should work fine in practice.

Parametrized types aren't supported because JSON Schema doesn't support them. We could support them by monomorphizing the types. This is provided by `atdcat -x` but it comes with complications which didn't seem worth solving just yet.

Closes #284

### PR checklist

### DIFF
--- a/.circleci/setup-system
+++ b/.circleci/setup-system
@@ -20,8 +20,8 @@ sudo apt-get install -y \
   python-is-python3 \
   scala
 
-# For atdpy
-pip install pytest mypy flake8
+# For JSON Schema and atdpy testing
+pip install jsonschema pytest mypy flake8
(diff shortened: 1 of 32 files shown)


mjambon:
I'm going to assume everyone is fine with this. It doesn't change existing features anyway. Next, I'll add support for `<doc ...`> annotations so as to include documentation in the JSON Schema translation.

Issue #273, opened 2022-04-08
Encode int64


AndreyErmilov:
Hi, 
After upgrading atdgen from version 2.2.1 to 2.3.3, I saw that all int64 numbers in the service response are converted to strings. There is issue #231 about it. That's okay, but there is a case where it doesn't fit. 
Is it possible to add a way to customize this behavior? For example, by default int64 numbers will remain numbers, but when adding a parameter they will be converted to strings.

AndreyErmilov:
I use timestamp with nanosecond accuracy. This is needed for tracing to work correctly (in my case it is zipkin). In order to express timestamp with nanoseconds I need to use int64.

AndreyErmilov:
I need to use Zipkin and Zipkin expects integers (timestamp field). I don't have the choice to convert int64 to string or not. If I had control over this behaviour it would be very useful.

Lupus:
Zipkin V2 API requires timestamp field to be integer: https://github.com/openzipkin/zipkin-api/blob/master/zipkin2-api.yaml#L319

```
    properties:
      timestamp:
        type: integer
        description: |
                    Epoch **microseconds** of this event.
                    For example, 1502787600000000 corresponds to 2017-08-15 09:00 UTC
                    This value should be set directly by instrumentation, using the most precise
                    value possible. For example, gettimeofday or multiplying epoch millis by 1000.
                    ```

Is there a way to express this in atd after #231 ? It does not fit in 32 bits.

[RFC 8259, Section 6](https://datatracker.ietf.org/doc/html/rfc8259#section-6) tells us that valid JSON allows arbitrary precision, and many implementations allow 64-bit values to be expressed natively, without string encoding.

**Comment by zindel** [comment]
I am more comfortable with keeping the string representation of `int64` by default. But I agree that there are cases when you need to have it be `int`.

How about we use the same trick as with `float`: https://atd.readthedocs.io/en/latest/atdgen.html#floats ?
E.g. for the example above:
```
type timestamp = int64 <repr="int">
```

**Comment by mjambon** [comment]
I'm confused. Sorry I didn't realize this earlier.

The following ATD definition:
```
type int64 = int <ocaml repr="int64">
```
means that in all the target languages except OCaml, the data will use the default representation of `int`. For JSON, it's a number, not a string. The current implementation of atdgen uses the function `Oj_run.write_int64` which writes a JSON string rather than a JSON number.

Valid ATD code for representing an OCaml int64 as a JSON string would be:
```
[...]

PR #291, opened 2022-05-06
Allow `<json repr="int">` for int64 types

zindel:
One way to address the concern in #273 (allow to encode `int64` value as `int` in JSON). Example:
```
$ cat x.atd && make && _build/default/atdgen/bin/ag_main.exe -j x.atd && cat x_j.ml
type i = int <ocaml repr="int64">
type i1 = int <ocaml repr="int64"><json repr="int">
...
type i1 = X_t.i1
type i = X_t.i
let write_i1 = (
  Atdgen_runtime.Oj_run.write_int64_as_int
)
...
let write_i = (
  Atdgen_runtime.Oj_run.write_int64
)
...
```
The pro is that we don't introduce new annotations and basically re-using the same `repr="int"` instruction as it is done for floats. On the other hand, there is some smell in it: another accepted value is `string`, but it is no-op for any int flavour, but `int64`.  Can be solved by using some specific `<json int64_as_int>` or something 🤷 Advice is welcomed :)
If we like the direction, I will complete the below checklist

cc @AndreyErmilov @Lupus
 
### PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+2.10.0 (XXXX-XX-XX)
+-------------------
+
+* atdgen: support `<json repr="int">` for `int64` values
+
 2.9.1 (2022-06-10)
 ------------------
 
--- a/atd/src/json.ml
+++ b/atd/src/json.ml
@@ -2,6 +2,10 @@
   Mapping from ATD to JSON
 *)
 
+type json_int =
+  | Int
+  | String
+
 type json_float =
   | Float of int option (* max decimal places *)
   | Int
@@ -57,7 +61,7 @@ type json_repr =
   | External
   | Field of json_field
   | Float of json_float
-  | Int
+  | Int of json_int
   | List of json_list
   | Nullable
   | Option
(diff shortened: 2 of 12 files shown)


mjambon (approved):
The code looks good to me in spirit. Just make to add tests, update the changelog, and update the atdgen documentation (i.e. check the boxes from the PR template).

Other target languages (e.g. python, scala etc.) won't support this feature just yet. It would be great if we could emit an error explaining that `int <json repr="string">` isn't supported, but I think it's fine for now.

mjambon:
Please see [my remark](https://github.com/ahrefs/atd/issues/273#issuecomment-1314323175). In essence, the ATD type defines the default representation for all target languages.
```
type i1 = int <ocaml repr="int64"> <json repr="int">
```
is redundant and should be written more simply as
```
type i1 = int <ocaml repr="int64">
```
Both forms must be equivalent since both imply that the JSON representation uses the usual representation for `int`, which is a JSON number (written without a decimal point like an integer is usually denoted in English).

mjambon:
I think we can close this now.

PR #292, opened 2022-05-06
fix `make all` when executed after `make test`

zindel:
Minor build-related fix. When I run `make test` and then `make` I get this:
```
$ make
dune build
Error: Multiple rules generated for
_build/default/atdpy/test/python-tests/everything.py:
- file present in source tree
- atdpy/test/python-tests/dune:4
-> required by alias default
Hint: rm -f atdpy/test/python-tests/everything.py
Error: Multiple rules generated for
_build/default/atdts/test/ts-tests/everything.ts:
- file present in source tree
- atdts/test/ts-tests/dune:4
-> required by alias default
Hint: rm -f atdts/test/ts-tests/everything.ts
make: *** [Makefile:12: all] Error 1
```

The fix blindly applies the hints from `dune` . If better fix is possible (within `atdts`/`atdpy`) it be great

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,8 @@ DUNE ?= dune
 # via opam.
 .PHONY: all
 all:
+	rm -f atdpy/test/python-tests/everything.py
+	rm -f atdts/test/ts-tests/everything.ts
 	$(DUNE) build
 
 # Install the OCaml dependencies for the build.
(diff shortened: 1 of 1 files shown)


mjambon:
Ah, yes, sorry about that. This fix should do it. Let me see if there's a slightly cleaner solution.

mjambon:
I made an edit to call `make clean-for-dune`. The whole thing is due to tensions between `dune` and `make`. I just prefer `make` for some things like testing and inspecting files.

mjambon (approved):

PR #295, opened 2022-05-12
Interpret 'abstract' as "any data" by default

mjambon:
This generalizes the use of the `abstract` predefined type to mean "any data" by default.

So far, `abstract` was supported only by atdgen (OCaml/Bucklescript backends) and limited to the following syntax:
```
type foo <...> = abstract
```
where:
* the right-hand side had to be be exactly `abstract`, not for example `abstract list`;
* the annotation `<...>` was required and would indicate which OCaml module and functions to use to convert the data.

Now, `abstract` can appear anywhere in a type expression e.g.
```
type mystery_things = abstract list
```
and the code generators should have a default interpretation of `abstract` that means "any data" and make a best effort to deal with it. For languages where the runtime representation of data is already some sort of AST such as TypeScript or Python+mypy, the type is `any` or `Any` and the result of parsing JSON needs no further transformation. For OCaml, however, we would need to parse the input into an explicit AST such as `Yojson.Safe.t` provided by the [yojson library](https://github.com/ocaml-community/yojson).

Note that the ATD type `abstract` translates to `Yojson.Safe.t` regardless of whether we're targeting JSON or biniou because both share the same type definitions. While we're not supporting this feature for biniou for now, it could be supported in the future. It could be done either by using the JSON AST to represent parsed biniou data or with the help of an annotation placed on `abstract` that would substitute `Yojson.Safe.t` by a type that's more appropriate for biniou.

This PR will implement this extended support for `abstract` in:
- [x] JSON Schema export with atdcat
- [x] Python (atdpy)
- [x] TypeScript (atdts)
- [x] OCaml ~~/Bucklescript~~ (atdgen)

The same work should be done for Java and Scala but it can be done later or when needed.

### PR checklist


aryx:
Thanks! great work.

mjambon:
This is ready for review!

tl;dr we can write `type things = abstract list` to represent a list of untyped JSON values.

aryx:
Can we merge this? This is really useful for languages like Python and Typescript!

mjambon:
I'll merge and release this tomorrow if there's no objection. Things to pay attention to:
* backward compatibility
* feature creep
I think we're good on both fronts. Anything else?

PR #296, opened 2022-05-23
Add `{!type:…}` annot in docs to help odoc with ambiguities

MisterDA:
It's possible that a type and a value share the same name in generated
code, and odoc can't know which of them the reference is pointing to.


mjambon (approved):
Looks great. Thank you, @MisterDA

PR #297, opened 2022-05-31
ATD modules

mjambon:
### PR checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,7 @@ js:
 clean:
 	$(DUNE) clean
 	$(MAKE) -C atdpy clean
+	$(MAKE) -C atdts clean
 	rm -rf tmp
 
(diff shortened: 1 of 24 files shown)


mjambon:
This attempt to support dedicated syntax for imports failed due to the complexity of the atdgen implementation. The problem is that atdgen generates code that doesn't read or write via a generic JSON AST but instead reads and writes JSON directly to/from bytes. This makes things unmanageable. I didn't have any problems supporting `import` for the other target languages (Python, TypeScript, ...) which all go through a JSON AST of sorts.

Here's what I suggest:

1. Salvage what's salvageable from this PR i.e. add support for `import` for all languages except OCaml/atdgen.
2. Implement a new, simpler OCaml code generator named [atdml](https://github.com/ahrefs/atd/issues/313) that would support imports.

PR #298, opened 2022-06-01
atdpy: Propagate decorators on sum types to all constructor classes

mmcqd:
Previously the following code would add `dataclass(frozen=True)` only to the class generated for `type t`, but _not_ the classes generated for its constructors. 
```
type t 
  <python decorator="dataclass(frozen=True)"> = [
  | A of int
  | B 
]
```

Now any decorators on the sum type class are propagated to the classes for its constructors.

This makes sense because any decorators applied to the sum type are meant to affect the inhabitants of the sum type, and in the generated python the inhabitants of a sum type are instances of the classes generated for its constructors.

### PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@ Next release
 ------------------
 
 * atdgen: use odoc syntax to disambiguate clashing names (#296).
+* atdpy: propagate decorators on sum types to all constructor classes
 
 2.7.0 (2022-05-17)
 ------------------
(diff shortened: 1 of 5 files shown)


mjambon:
Great, thank you! We should add a sentence in the documentation (https://atd.readthedocs.io/en/latest/atdpy.html#custom-class-decorators). `CONTRIBUTING.md` explains how to preview locally. If it's just English text without markup, you can just edit the source file in `/doc` and assume it's going to be fine.

mjambon (approved):

PR #299, opened 2022-06-03
Adapt to use Yojson 2.0 API

Leonidas-from-XIV:
Yojson 2.0 removed the `Stream` API to be forward-compatible with OCaml 5 (and stop triggering deprecation warnings in OCaml 4.14). To make atd use Yojson 2.0 going forward this PR modifies the generated code + the runtime to use the new functions. It comes with the small advantage of not needing to use the `Stream`-compatibility package anymore.

Older versions have been restricted to use Yojson < 2.0: https://github.com/ocaml/opam-repository/pull/21464

### PR checklist

- [X] Documentation is up-to-date


Leonidas-from-XIV:
We're gonna improve the pretty-printing a bit in Yojson 2.0.1: https://github.com/ocaml-community/yojson/pull/141

mjambon:
Thanks for that work @Leonidas-from-XIV. I think it's best to wait for yojson 2.0.1 before merging this. The closest we're to the original output, the better.

Leonidas-from-XIV:
@mjambon I completely agree, I'll rebase this PR when we have 2.0.1 in place. Currently wrestling with getting 2.0.0 into OPAM without too much breakage but from there on, 2.0.1 should be smooth sailing.

Leonidas-from-XIV:
Updated to use 2.0.1 which prettyprints the same way as the older release so no changes necessary, removed that commit altogether.

mjambon:
Thank you for the update. This looks good to me. I'm on vacation without a
laptop for another 10 days so we can merge it but someone else would have
to make the release or any unexpected fixes.

mjambon (approved):
Great, thanks again.

mjambon:
I didn't notice that `make test` wasn't working before merging. I'm going to do something to revert this merge until we have a new yojson release that includes the fix https://github.com/ocaml-community/yojson/pull/150

mjambon:
This PR was resurrected as https://github.com/ahrefs/atd/pull/311

Issue #263, opened 2022-03-23
Don't validate abstract `Yojson.Safe.t` abstract type


MisterDA:
Yojson rename it's `Yojson.Safe.json` type to `Yojson.Safe.t` and deprecated the old type. It also removed the `Yojson.Safe.validate_json` and deliberately not introduced a new `Yojson.Safe.validate_t`. I have an atd document using the json type, and I'd like to validation for the values produced, but however I write it, I cannot get atd to turn off validation for that type: it always produces code that call the non-existent `Yojson.Safe.validate_t` function.

```shell
cat > test.atd <<EOF
type json <ocaml module="Yojson.Safe" t="t"> = abstract <ocaml valid="fun _x -> true">
type json' <ocaml module="Yojson.Safe" t="t"> = abstract <ocaml validator="fun _x _y -> None">
EOF
atdgen -v test.atd
ocamlfind ocamlc -o test.exe -linkpkg -package yojson,atdgen test_v.ml
```
produces
```ocaml
(* Auto-generated from "test.atd" *)
              [@@@ocaml.warning "-27-32-35-39"]

type json' = Yojson.Safe.t

type json = Yojson.Safe.t

let validate_json' = (
  (fun path x -> match ( fun _x _y -> None ) path x with | Some _ as err -> err | None -> (Yojson.Safe.validate_t) path x)
)
let validate_json = (
  (fun path x -> match ( fun path x ->
    let msg = "Failed check by fun _x -> true" in
    if (fun _x -> true) x then
      None
    else
      Some (Atdgen_runtime.Util.Validation.error ~msg path) ) path x with | Some _ as err -> err | None -> (Yojson.Safe.validate_t) path x)
)
```
and
```
File "test_v.ml", line 9, characters 91-113:
9 |   (fun path x -> match ( fun _x _y -> None ) path x with | Some _ as err -> err | None -> (Yojson.Safe.validate_t) path x)
                                                                                               ^^^^^^^^^^^^^^^^^^^^^^
Error: Unbound value Yojson.Safe.validate_t
```

mjambon:
I think [this change in yojson](https://github.com/ocaml-community/yojson/pull/137) is going to break user code no matter what we fix in atdgen, because the user code is `type json <ocaml module="Yojson.Safe" t="t">` in their ATD file. I don't think atdgen should interpret `Yojson.Safe` in a special manner either because it's already complicated enough as it is.

Some ideas:
[...]

PR #301, opened 2022-06-07
atdgen: update `abstract` type validation to accept all input by default

metanivek:
This PR follows up on #295 and #263 by making the default validation for `abstract` types pass for any input, while still allowing for custom validation.

I used the following definition for testing, which appears to my eyes to be correct and compiles successfully (with a small `util.ml` with always-pass validation definitions).
```
type raw_json = abstract
type raw_json_valid = abstract <ocaml valid="fun _ -> true">

type demo = {
  one: raw_json;
  two: raw_json <ocaml validator="Util.validate_abstract">;
  three: raw_json <ocaml valid="fun _ -> true">;
  four: abstract;
  five: abstract <ocaml validator="Util.validate_abstract">;
  six: abstract <ocaml valid="fun _ -> true">;
  seven: raw_json_valid;
}
```

Feedback very much welcome since this is my first time in this code base!

### PR checklist

### DIFF
--- a/atdgen/src/ov_mapping.ml
+++ b/atdgen/src/ov_mapping.ml
@@ -207,6 +207,8 @@ let rec mapping_of_expr
            Float (loc, Float, (v an, true))
        | "string" ->
            String (loc, String, (v an, true))
+       | "abstract" ->
+           Abstract (loc, Abstract, (v an, true))
        | s ->
            let validator =
              match v2 an x0 with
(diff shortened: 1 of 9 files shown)


atdgen/src/ov_mapping.ml
```diff
@@ -307,13 +309,14 @@ let def_of_atd is_shallow (loc, (name, param, an), x) =
...
     match as_abstract x with
     | Some (_, an2) ->
-        Ocaml.get_ocaml_module_and_t Validate name an
-        |> Option.map (fun (types_module, main_module, ext_name) ->
-          let args = List.map (fun s -> Tvar (loc, s)) param in
-          External (loc, name, args,
-                    Ocaml.Repr.External (types_module, main_module, ext_name),
-                    (Validate.get_validator an2, false))
-        )
+        (match Ocaml.get_ocaml_module_and_t Validate name an with
```
metanivek, line 312: The best I can tell, there seem to be two notions of "abstract" in the code base. This change allows for `type raw_json = abstract` to work as expected while (I think) still respecting the existing `as_abstract` notion.

mjambon:
Hi! Personally, I haven't done much with validators in a while, so it's not something that was on my mind when I recently extended what we can do with `abstract`. It would be great if you could post an example illustrating the problem you're solving (input, output, expected output). After that, you should add some tests that show that your fix works.

mjambon:
After looking at the tests we have, it looks indeed that we have nothing for testing the validation of `abstract` values. There's the file `test_abstract.atd` which could be reused for testing validation, or you can create a new one just for this. I think it's nice to keep test files focused on one feature rather than putting everything in a giant `test.atd` like we still have (note that you may not be able to add `type foo = abstract` to `test.atd` due to the biniou backend not supporting it). You'll probably have to edit `atdgen/test/dune` to add instructions for running `atdgen -v`.

metanivek:
@mjambon just pushed added/updated tests. If you think it would be good to have tests for the validators (like in my examples), let me know. I'm happy to add that as well.

mjambon:
The tests you added are sufficient, I think. Thank you! I'll let you update the changelog.

metanivek:
@mjambon changelog updated (and title of PR :)

mjambon:
Thank you! Merging and starting a release.

PR #302, opened 2022-06-08
[atdts] fix case reader

gares:
We are using atdgen and atdts in LPCIC/elpi#146 and we found out that the ts reader is just broken, probably a copy/paste error.
Here the fix.

### PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+UNRELEASED
+------------------
+
+* atdts: fix reader for case type (#302)
+
 2.8.0 (2022-06-06)
 ------------------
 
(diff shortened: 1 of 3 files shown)


mjambon:
atdts is still very new. Thanks for reporting and fixing this. Is there a test that would have caught the problem?

gares:
you should write a .ts file which compses reading and writing and checks the composition is the identity.

anyway, the output after this pr is ok for us, here the atd file: https://github.com/LPCIC/elpi/pull/146/files?file-filters%5B%5D=.atd&file-filters%5B%5D=.mly&file-filters%5B%5D=.opam&file-filters%5B%5D=.ts&file-filters%5B%5D=.yml&file-filters%5B%5D=No+extension&show-viewed-files=true#diff-2eab66a8811c8c40f2c50e00de4842d5b989fd2d43e63aade37acb9e72b98886

mjambon:
ok, let's merge this.

PR #315, opened 2022-08-09
Allow recursive references in atdpy

zyannes:
Use Python's future annotations to support forward declarations so that atdpy can generate classes with recursive references.

### PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,10 +2,6 @@ Master
 ------------------
 
 * atdpy: Support recursive definitions
-
-Master
-------------------
-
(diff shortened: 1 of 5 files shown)


mjambon:
Neat. Can you please add a test? For that, you would normally edit https://github.com/ahrefs/atd/blob/master/atdpy/test/atd-input/everything.atd and https://github.com/ahrefs/atd/blob/master/atdpy/test/python-tests/test_atdpy.py

A comment explaining why we need `from __future__ import annotations` would be great too.

mjambon (approved):

Issue #303, opened 2022-06-11
[atdpy/atdts] support for polymorphic types


aryx:
See https://github.com/returntocorp/semgrep-interfaces/blob/main/ast_generic_v0.atd
and specially:
type 'a wrap_ = ('a * tok)

type 'a bracket = (tok * 'a * tok)

A simple implementation would be to inline the definition at use time ...

mjambon:
The `atd` library provides an [option to monomorphize](https://github.com/ahrefs/atd/blob/110d079dac5a3babfa2af69887ee8992cb5e213c/atd/src/util.mli#L24) all type definitions. This may come in handy.

PR #319, opened 2022-11-10
Add support for parametrized types to atdpy

mjambon:
Implements the atdpy part of https://github.com/ahrefs/atd/issues/303

There were annoyances due to the representation of lists, options and other parametrized built-ins which sometimes use the dedicated `List`, `Option` etc. constructors and other times use the generic `Name` constructor. I solved it by providing functions to convert between the two forms.

### PR checklist

### DIFF
--- a/atdpy/test/python-expected/everything.py
+++ b/atdpy/test/python-expected/everything.py
@@ -407,6 +407,58 @@ class Alias:
 
 
 @dataclass
+class X2:
+    """Original type: _2"""
+
+    value: Tuple[Kind, Kind, int]
+
+    @classmethod
+    def from_json(cls, x: Any) -> 'X2':
+        return cls((lambda x: (Kind.from_json(x[0]), Kind.from_json(x[1]), _atd_read_int(x[2])) if isinstance(x, list) and len(x) == 3 else _atd_bad_json('array of length 3', x))(x))
+
+    def to_json(self) -> Any:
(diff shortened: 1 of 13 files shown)


atdpy/test/python-expected/everything.py
```diff
@@ -406,6 +406,58 @@ def to_json_string(self, **kw: Any) -> str:
...
+        return res
+
+    @classmethod
+    def from_json_string(cls, x: str) -> 'X1':
+        return cls.from_json(json.loads(x))
+
+    def to_json_string(self, **kw: Any) -> str:
+        return json.dumps(self.to_json(), **kw)
+
+
```
mjambon, line 460: These "anonymous" classes `X1` and `X2` aren't great at all for the Python user. We could get rid of `X1` by inlining the tuples. We'd still have anonymous classes for monomorphized records and variants. If we stick with monomorphization, an idea to improve the situation is to generate better names than `X1`, `X2`, etc.

In [our AST definition](https://github.com/returntocorp/semgrep-interfaces/blob/main/ast_generic_v1.atd), we have:
```ocaml
type 'a wrap_ = ('a * tok)
```

It occurs as `string wrap_`, `int wrap_` etc. for which numbered types `_1`, `_2` etc. are created as the result of monomorphization (`_1` becomes Python class `X1`). Instead of numbered type names, we could generate better names like `string_wrap_` and `int_wrap_`.

The nice thing about monomorphization is that it's done once and for all, benefiting all the target languages. Alternatively, if we drop monomorphization, we'd  have to create polymorphic Python classes and involve higher-order functions like we already do for lists and nullables.

PR #320, opened 2022-11-10
Support option types in Python by treating them as nullables

mjambon:
Note that 'Some None' of type 'int option option' can't be represented with nullables but the problem hopefully won't come up frequently.

### PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,5 @@
 * atdpy: Support parametrized type definitions (#303)
+* atdpy: Support options approximately by treating them as nullables
 
 2.10.0 (2022-08-09)
 -------------------
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,5 @@
 * atdpy: Support parametrized type definitions (#303)
-* atdpy: Support options approximately by treating them as nullables
+* atdpy: Support options approximately by treating them as nullables (#320)
 
 2.10.0 (2022-08-09)
 -------------------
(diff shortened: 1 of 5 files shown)


Khady (approved):
having different behaviors between languages for the same type of value is a bit surprising. But I'm not sure everything else is coherent :shrug:

mjambon:
The problem we were trying to solve is to support a [big ATD file](https://github.com/returntocorp/semgrep-interfaces/blob/main/semgrep_output_v1.atd) that we wrote, which represents an AST. In this context, `option` and `nullable` could be used interchangeably.

The solutions I considered were:

a. Writing atd files to use `nullable` instead of `option`. Both translate to the OCaml `option` type already.
b. Creating our own `Option` class in Python and using that.
c. Adding a flag to atdpy telling it to treat option as nullable.
d. Having atdpy always treat `option` as `nullable`.

This PR implements (d) because it requires no special effort for Python users or for the authors of the ATD interfaces. It's also easy for us to add to atdpy. The only downside I see is that nested option types like `int option option` won't work. I think such types are pretty bad anyway. They might occur in OCaml on occasion but seem like a confusing way of defining a public interface. When this happens, Python users should complain and have the ATD-backed API changed. If there's ever a scenario in which such a change is impossible, we could also modify atdpy and add support for (c'):

c'. [= b+c] Adding a flag to atdpy telling it to _not_ treat option as nullable and instead use a dedicated `Option` class.

mjambon:
This behavior is incompatible with atdgen as @Khady already pointed out. This is causing us problems right now. I'm working on undoing it. See https://github.com/ahrefs/atd/issues/332

PR #321, opened 2022-11-11
Assign stable and human-friendly names to generated type names

mjambon:
This is a follow-up to yesterday's PR on which I commented: https://github.com/ahrefs/atd/pull/319#pullrequestreview-1176225840

The nicer type names are used:

* internally in `.ml` files generated by atdgen. The name changes shouldn't affect the users since they're not part of the generated .mli` interfaces.
* for Python class names resulting from monomorphization. Check out the changes in the generated file `everything.py`.

### PR checklist

### DIFF
--- a/atdgen/test/test.expected.ml
+++ b/atdgen/test/test.expected.ml
@@ -166,33 +166,33 @@ type 'a abs2 = 'a list
 
 type 'a abs1 = 'a list
 
-let _19_tag = Bi_io.array_tag
-let write_untagged__19 _a_tag write_untagged__a write__a = (
+let _a_list_tag = Bi_io.array_tag
+let write_untagged__a_list _a_tag write_untagged__a write__a = (
   Atdgen_runtime.Ob_run.write_untagged_list
     _a_tag
     (
       write_untagged__a
     )
 )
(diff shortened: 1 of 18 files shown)


atdgen/test/test.expected.ml
```diff
@@ -1369,82 +1423,28 @@ let get__4_reader = (
           | 0x80 ->
             Some (
               (
-                Atdgen_runtime.Ob_run.read_int
+                Atdgen_runtime.Ob_run.read_bool
```
mjambon, line 1426: This looks wrong.

atdgen/test/testj.expected.ml
```diff
@@ -1551,7 +1583,7 @@ let read__6 = (
             | "Some" ->
               Atdgen_runtime.Oj_run.read_until_field_value p lb;
               let x = (
-                  Atdgen_runtime.Oj_run.read_string
+                  Atdgen_runtime.Oj_run.read_int
```
mjambon, line 1586: This also looks wrong. Could be a bad diff too.

atdgen/test/test.expected.ml
mjambon: nah, it's a bad diff.

PR #333, opened 2023-03-14
atdpy: fix JSON representation of the option type

### LINKED ISSUE #332: atdpy treats option as nullable
atdpy treats the `option` type as `nullable` but it should not.

The convention used by ATD is to use `"None"` or `["Some", 42]` to represent `int option` values in JSON. Python code generated by atdpy instead expects `null` or `42` as if the type was `int nullable` rather than `int option`.

I don't think we need to support the `option` type at all for Python since Python doesn't have a standard option type. The `Optional` type known by mypy is in fact equivalent to ATD's `nullable`, not `option`.

Instead of emitting an error, we could create an option type for Python but it would probably be awkward. This can be done later if there's some demand for it (I doubt it will be the case).

mjambon:
Fixes #332

[Previously](https://github.com/ahrefs/atd/pull/320), I added Python "support" for the option type by making it behave like nullable. While this is convenient on the Python programmer's side, it's incompatible with the ATD convention for representing the option type in JSON.

The solution I'm adopting here is:
* keep using mypy's `Optional` type which is a nullable type;
* adopt the conventional JSON representation e.g. `"None"` or `["Some", 42]` instead of `null` or `42`, respectively.

Note that ATD's `nullable` type is still recommended to make JSON code more idiomatic and more readable when it's possible. The only problematic case is with nested nullables since Python `null` ambiguously represents JSON `"None"` and JSON `["Some", "None"]`. Because of these issues, authors of APIs are advised to not use nested nullables or nested options.

### PR checklist


Khady (approved):
LGTM

PR #355, opened 2023-10-03
[ATDD] use struct with `alias this` instead of `Typedef` for user defined types

elrandar:
This PR fixes various issues with the interoperability of user defined types, used in or outside of records. It makes use or the [alias this](https://dlang.org/blog/2018/05/21/complicated-types-prefer-alias-this-over-alias-for-easier-to-read-error-messages/) Dlang feature, instead of the `Typedef` that was used previously. 
This allows for alias type to be defined as structs and used in the same way as the type they wrap, while still maintaining different behaviours for differently wrapped types.


atdd/src/lib/Codegen.ml
```diff
@@ -510,7 +510,7 @@ auto toJsonString(T)(T obj)
   |}
     atd_filename
     atd_filename
-    (Filename.remove_extension atd_filename)
+    (sprintf "%s_atd" (Filename.remove_extension atd_filename))
```
Khady, line 513: I'm guessing this is a breaking change, but given that atdd wasn't officially announced it shouldn't be a real problem

Khady (approved):

atdd/src/lib/Codegen.ml
mjambon: oops, I started release 2.13.0 without merging this PR. I'm planning to make another release in a few days.

mjambon: I added a changelog entry for this PR. Please fix it as needed. https://github.com/ahrefs/atd/pull/355/commits/2f4418c19c0f2e5460cb756cdfc072c609f59efe

Issue #352, opened 2023-09-29
Provide a tool to compare two ATD files for compatibility


mjambon:
@jbergler wrote:
> To detect breaking changes [in an ATD file] I currently do something like this and manually review the changes
```
~/co/semgrep-interfaces % git diff v1.17.0..main *.atd
```
> But these diffs are huge, so I don't really get much confidence out of asking a human to review the changes.
> My past experience tells me the only way we get high confidence outcomes is to automate the review for these kinds of nuanced details.
> It's ok if we want to say we don't think the effort to get there is required. Before we do that I want to brainstorm how we might achieve that.
> In protobuf land, there are cli tools that compare two proto files and print all the breaking changes. Could you have a guess at how much work it would be to build something similar for atd?

I think it would be a relatively low effort to produce a tool (`atddiff`?) that compares two ATD files and reports differences:
* incompatibilities in one direction (left: writer, right: reader)
* incompatibilities in the other direction (left: reader, right: writer)
* `<`...`>` annotations that changed (but without trying to interpret them)
* bonus: track renames and ignore name changes that don't affect data compatibility

For example, let's compare two files, `Interface_v1.atd` against `Interface_v2.atd`:
```ocaml
(* Interface, version 1 *)

type fruit = [ Kiwi | Strawberry ]

type veggie = [ Potato | Eggplant ]

type basket = {
  owner: string;
  fruits: fruit list;
  veggies: veggie list;
  ?notes: string option;
}
```

```ocaml
(* Interface, version 2 *)

type fruit = [ Kiwi | Strawberry | Plum ]

type vegetable = [ Eggplant | Potato ]

type basket = {
  fruits: fruit list;
  ~veggies: vegetable list;
  notes: string;
  price: int;
}

type drink = [ Water | Unicorn_blood ]
```

```
$ atddiff Interface_v1.atd Interface_v2.atd
Backward incompatibilities result in the following problems:
- cannot read old data using new implementation
- cannot read responses from old server in new client
- cannot read requests from old client in new server
Forward incompatibilities result in the following problems:
- cannot read new data using old implementation
- cannot read responses from new server in old client
- cannot read requests from new client in old server

Old file Interface_v1.atd, line 123,
New file Interface_v2.atd, line 321:
Backward incompatibility: the 'notes' field is required by newer implementations but used to be optional.

Old file Interface_v1.atd, line 123,
New file Interface_v2.atd, line 321:
Backward incompatibility: the new 'price' field is required by newer implementations.

[...]

PR #358, opened 2023-10-12
new: atddiff tool

mjambon:
This implements #352 

~~I still need to review the test results more carefully and add tests.~~ For now, the output is text. We can add JSON output in another PR (see https://github.com/ahrefs/atd/issues/360).

### PR checklist


mjambon:
to do: avoid infinite looping on recursive type definitions

atddiff/src/bin/Atddiff_main.ml
```diff
@@ -0,0 +1,201 @@
...
+  Arg.value (Arg.opt (Arg.some Arg.string) None info)
+
+let json_defaults_term =
+  let info =
+    Arg.info ["json-defaults"]
+      ~doc:"Shorthand for '--json-defaults-old --json-defaults-new'."
+  in
+  Arg.value (Arg.flag info)
+
+let json_defaults_old_term =
```
aryx, line 76: I like to put the type on those things, like let json_defaults_old_term : bool Term.t = ...
It helps a bit.

atddiff/test/test_new.atd
```diff
@@ -0,0 +1,75 @@
+(*
```
aryx, line 1: Ideally this would split in different test files for each scenario and each file could contain 3 sections:
old_atd content
new_atd_content
expected output of atdiff

so you see everything in one file instead of now you need to inspect 3 separate files to understand.
But not super important. More important to get v0 out of the door!

aryx (commented):
Nice!

atddiff/test/test_new.atd
mjambon: It turns out dune doesn't support dynamic targets. I created a shell script that gathers test files and generates a `dune` file. I wish there was a simpler and cleaner way.

Anyway, now it's easy to split the test files.

aryx:
ship it!

atddiff/src/bin/Atddiff_main.ml
mjambon: I did what you suggest. I like it.

mjambon:
This is feature-complete for a first release unless I forgot something. I'll update the documentation in another PR while the opam release is in progress.

PR #359, opened 2023-10-15
Remove unused code for supporting '<json lowercase_tags>'

mjambon:
`<json lowercase_tags>` was meant to be supported but I see that none of the code generators consult it, so I'm removing the support we have for it from our `Atd.Json` module.

This is an internal change so I don't think it deserves an entry in the changelog. Let me know if you think otherwise.

### PR checklist


mjambon:
Merging. If this breaks some external dev's code, I'd love to know who they are.

PR #375, opened 2023-10-23
Migrate from Bucklescript to Melange

jchavarri:
BuckleScript [is no more](https://rescript-lang.org/blog/bucklescript-is-rebranding). Its legacy lives through either [ReScript](https://rescript-lang.org/) or [Melange](https://melange.re/).

This PR migrates all related code to Melange, as that is the compiler we continue using at Ahrefs. This involves:
- Rename `-bs` flag to `-mel`
- Generated files suffix migrated from `_bs` to `_mel`
- Rename internal code and descriptions in package definitions

Additionally, I included myself as maintainer of this part of the code.

### PR checklist

- [x] There is no new code, so no need to add tests to catch future regressions


atdgen/src/ocaml.ml
```diff
@@ -134,7 +134,7 @@ let path_of_target (target : target) =
     | Default -> [ "ocaml" ]
     | Biniou -> [ "ocaml_biniou"; "ocaml" ]
     | Json -> [ "ocaml_json"; "ocaml" ]
-    | Bucklescript -> ["ocaml_bs"; "ocaml"]
+    | Melange -> ["ocaml_mel"; "ocaml"]
```
cyberhuman, line 137: I think we want to support the old annotations for backwards compatibility (maybe emitting a warning that the old name will eventually be removed?)

jchavarri: I didn't know there was this kind of annotation available. I don't understand well what is the purpose?

They are not documented or tested anywhere in the atd codebase, and at least Ahrefs doesn't use it (afaict), so I am not sure it is worth to spend time providing a smooth migration experience.

cyberhuman: `ocaml` applies to all generators but `ocaml_xxx` only to `xxx`, and the latter have a priority.

jchavarri: Hm, I see. I was expecting atdgen to complain if an unknown generator was used, but i made some tests and observed that doesn't seem to be the case, it just ignores them.

Do you think it's worth to solve the migration for the specific `ocaml_bs` case? Maybe it's more valuable to solve the common case and have atdgen error out if an unknown generator is used?

jchavarri: this is what i'm testing fwiw

```
type 'a t <ocaml_foo module="Abs"> = abstract
```

cyberhuman: I think solving the general case is better. It's better to error out straight away if a tag is not supported, and it will cover the `ocaml_bs` case automatically.

cyberhuman: Actually now thinking about it, it may be tough since we share generators with the atdgen-www fork.

jchavarri: For now, added back the `ocaml_bs` path in https://github.com/ahrefs/atd/pull/375/commits/f94ecc96b39f19ac189cf42469df504b6072d839.

Khady:
LGTM. It is going to break some backward compatibility, but that's probably fine?

jchavarri:
True, it is a breaking change, I updated the changelog entry to reflect it. I wouldn't bump a major version though.

@mjambon do you have any thoughts on the above and/or see any blockers before this PR can be merged?

jchavarri:
If no one disagrees, as @Khady already approved, I will merge this tomorrow.

Khady:
it looks like this PR is only doing some renaming, so maybe it's still possible to support both melange and rescript at the same time?

Btw I don't think there's an opposition to the support of rescript in atd. It's just that ahrefs isn't using rescript so there was no need to keep that feature alive. If someone bothers opening a PR to add the support again I guess that it could get merged? I don't know how useful atd would be for the rescript community.

TheSpyder:
I don't know how useful it would be to the wider community either. If it's just me I wouldn't worry too much.

Once I get the runtime and binary releases done for rescript, I'll post about it on the forum and see if there is any interest 🤔

TheSpyder:
It turns out the API border between `atdgen` and the runtime API is fundamentally incompatible with ReScript 11. They're pushing the new "uncurried" mode fairly hard, and while I managed to convert the API project fairly easily the generated code fails to compile because it makes use of curried functions.

I can and will continue; I was already planning to disable "uncurried" in my project. But it means community use is likely to be low unless I contribute a rescript-specific code generator path to `atdgen`. Whether that's worth it for my team is something we'll evaluate when (and if) we look at switching the project to "uncurried" mode.

Happy to chat more over email if you want, rather than continue in this old closed PR thread 🙂

jchavarri:
> Happy to chat more over email if you want, rather than continue in this old closed PR thread

I think there are upsides to keep the conversation public. Feel free to open a new issue in this repo to track and discuss the details of this addition.

Issue #393, opened 2023-11-18
atdd: workaround dlang compiler bug with SumTypes


elrandar:
Implement `SumTypes` using `alias this` in structs instead of `alias` in order to workaround `dlang` compiler bug.

PR #394, opened 2023-11-20
atdd: use `alias this` for SumTypes

elrandar:
- Fixes issue #393.
- Improves construction of Tuple types (see tests)


mjambon (approved):

Khady:
@elrandar was the discussion which lead to that fix public? If yes then it would be cool to add a link to it. Otherwise we'll merge as is.

PR #395, opened 2023-11-26
Fix small typo in data validation tutorial

mbarbin:
Hello,

I've spotted a minor code inconsistency in `atdgen-tutorial.rst`, in the [data-validation](https://atd.readthedocs.io/en/latest/atdgen.html#data-validation) section. The test for validation was reversed, `Some _` should indicate some errors were found, and `None` means validation passed (`is_valid = true`).

I've updated the test file to align with the code quoted in the doc, also changed a date which was out of sync (unimportant).

Thank you!

### PR checklist

- [x] New code has tests to catch future regressions
  Not applicable, no new code.

- [ ] Documentation is up-to-date
  Successfully ran `dune build @doc`. Any other checks required? 

- [x] `CHANGES.md` is up-to-date
  Not applicable.

PR #396, opened 2023-11-27
Add test to exercise validation tutorial demo

mbarbin:
Hello,

In this PR, I've proposed running the tutorial demo as part of `dune runtest` to ensure future-proofing. This change was prompted by the fix I made in test #395 (thanks for merging!).

If you find this approach beneficial, I'm more than willing to apply similar changes to the other tutorials. Assuming these would involve adding a few `dune` and `*.t` files, I'd be happy to contribute further.

I've noticed that adding `(cram enable)` to `dune-project` was necessary, which I believe is the default in a later version of `dune-lang`. If you're not comfortable with enabling cram tests globally, but still want diff tests in the tutorial, we can store the output in a file and use a dune diff rule instead. Please let me know your preference.

Thank you for considering my changes and for your work on `atd`.

PR checklist

This PR adds a new test that is run as part of `dune runtest`.
This PR does not involve changes that require updates to the documentation.
- [x] CHANGES.md is up-to-date
This PR does not involve changes that require updates to CHANGES.md.

### DIFF
--- a/doc/atdgen-tutorial-data/validate/dune
+++ b/doc/atdgen-tutorial-data/validate/dune
@@ -1,0 +1,24 @@
+(executable
+ (name resume)
+ (libraries atdgen-runtime yojson))
+
+(rule
+ (targets resume_t.ml resume_t.mli)
+ (deps resume.atd)
+ (action
+  (run %{bin:atdgen} %{deps} -t)))
+
+(rule
+ (targets resume_j.ml resume_j.mli)
+ (deps resume.atd)
+ (action
+  (run %{bin:atdgen} %{deps} -j)))
+
+(rule
+ (targets resume_v.ml resume_v.mli)
+ (deps resume.atd)
+ (action
+  (run %{bin:atdgen} %{deps} -v)))
+
+(cram
+ (deps resume.exe))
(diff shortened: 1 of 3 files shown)


mjambon:
Hi. I'm deeply sorry I missed the original GitHub notification.
This looks useful. I'm not familiar with cram tests yet and I'm not sure about the other project contributors. Don't hesitate to add comments about context and any questions that may arise ("what's this for?", "why is it not using [legacy solution]?", ...).

mbarbin:
Hi Martin,

I appreciate your time in reviewing this PR. Thank you for merging it!

I've been exploring atd as a potential solution for persisting data from an OCaml program in a human-friendly format. I've just started using atd last week, and so far, it looks like a good fit for my needs. Thank you for atd!

During my exploration of atd, I've been going through the documentation, tutorials, and experimenting with building things using dune. I've found that having the demos built with dune enhances my learning experience, as I can see the types within my editor using Merlin. This will also ensure that the demos are checked during your build process and kept up to date. That's what led me to create this PR.

I've updated the branch to include other tutorials as well. Please let me know if there's anything else you'd like me to add or modify.

Thanks again for your work on atd!

PR #401, opened 2024-02-08
atdgen: add support for open_enum in melange

tatchi:
This is my attempt at adding support for `<json open_enum>` in melange


CHANGES.md
```diff
@@ -4,6 +4,7 @@ Unreleased
 * atdgen: Breaking change, migrate from Bucklescript to Melange (#375)
 * atdd: Workaround d compiler bug regarding declaration order when using aliases (#393)
         Algebraic data types (SumType) now uses `alias this` syntax.
+* atdgen: Add support for `<json open_enum>` in melnage (#401)
```
jchavarri, line 7:
```suggestion
* atdgen: Add support for `<json open_enum>` in Melange (#401)
```

atdgen/test/melange/melangespec.atd
```diff
@@ -108,3 +108,15 @@ type variant2 = [
   | A
   | C
 ] <ocaml repr="classic">
+
+type with_open_enum = [
```
jchavarri, line 112: In the OCaml side, there's a test with a list + enum inside: 

https://github.com/ahrefs/atd/blob/77a07ed1f184f370d95ff882fa3a31911eecdf0f/atdgen/test/test3j.atd#L104

Do you think it's worth adding it to the melange suite as well?

atdgen/src/omelange_emit.ml
```diff
@@ -577,6 +591,12 @@ and make_sum_writer ?type_annot (p : param)
                 ; Line (sprintf "%s %s" (encoder_ident "constr0")
                           (make_json_string json_cons))
                 ]
+            | Some v when open_enum ->
+                (* v should resolve to type string. *)
+                [ Line (sprintf "| %s%s x ->" tick ocaml_cons)
+                ; Inline (make_writer p v)
```
jchavarri, line 597: Shouldn't this be inside a block? I'm taking the other impl as reference:

https://github.com/ahrefs/atd/blob/77a07ed1f184f370d95ff882fa3a31911eecdf0f/atdgen/src/oj_emit.ml#L436

Maybe would be helpful to review to have an overview on the things that changed from native impl?

jchavarri (commented):
Thanks! It looks good to me, but my understanding of this part of atd is very slow, would be great if someone else with more knowledge could take a look.

I saw the additions to `omelange_emit` and the generated code are essentially copied from the native part? If so, I wonder if there's an opportunity to share more code across both implementations in the form of a new library, or something else.

atdgen/test/melange/melangespec.atd
tatchi: done

atdgen/src/omelange_emit.ml
tatchi: I don't know, that's what I used first, but the indentation feels worse:

```diff
diff --git a/atdgen/test/melange/melangespec_mel.expected.ml b/atdgen/test/melange/melangespec_mel.expected.ml
index 744d15f..35f4242 100644
--- a/atdgen/test/melange/melangespec_mel.expected.ml
+++ b/atdgen/test/melange/melangespec_mel.expected.ml
@@ -170,7 +170,7 @@ let write_with_open_enum = (
     | `Beta ->
     Atdgen_codec_runtime.Encode.constr0 "Beta"
     | `Other x ->
-    Atdgen_codec_runtime.Encode.string
+      Atdgen_codec_runtime.Encode.string
     x
   )
 )
@@ -205,7 +205,7 @@ let write_variant3 = (
     | B ->
     Atdgen_codec_runtime.Encode.constr0 "B"
     | A x ->
-    Atdgen_codec_runtime.Encode.string
+      Atdgen_codec_runtime.Encode.string
     x
   )
 )
```

Khady:
what's the status here?

tatchi:
> I saw the additions to `omelange_emit` and the generated code are essentially copied from the native part? If so, I wonder if there's an opportunity to share more code across both implementations in the form of a new library, or something else.

I took inspiration from the native part, but I'm not sure if we can really share pieces of code. Anything particular you had in mind ?

> what's the status here?

I think I've addressed all of @jchavarri's comments, so I think it should be good to merge.

jchavarri (approved):
Thanks!

PR #403, opened 2024-03-01
atdgen: support for wrapped keys for object repr in melange

danielmercier:
This PR implements support for wrapped keys when using `<json repr="object">`:

As a simple example:
```ocaml
type t =
   ( string wrap <ocaml t="int" wrap="int_of_string" unwrap="string_of_int">
   * int) list <json repr="object">
```

Really unsure if this is the correct approach as re-using the encoder/decoder for the name is the simplest solution I found. This emulates what is done in `oj_emit.ml`. But this changes the Encode/Decode APIs

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@ Unreleased
 * atdgen: Breaking change, migrate from Bucklescript to Melange (#375)
 * atdd: Workaround d compiler bug regarding declaration order when using aliases (#393)
         Algebraic data types (SumType) now uses `alias this` syntax.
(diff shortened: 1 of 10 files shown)


danielmercier:
> This seems like an unusual need. Wouldn't it be best served by using an [adapter](https://atd.readthedocs.io/en/latest/atdgen.html#field-adapter-ocaml) that rewrites the JSON object into an array of pairs? I'm worried about the maintenance burden, atdgen being already very complicated.

This definitely sounds like a good use for the adapters. Agreed for the maintenance burden, I'll close this PR, thanks for the suggestion!

Issue #324, opened 2022-12-12
using a different ATD annotation tag name for rescript (and mélange?)


tsnobip:
Today, ocaml and rescript/mélange share the same ATD annotation tag name: `<ocaml>` when using the `-bs` option.

But this is becoming quite a pain because we share ATD files quite often between "native" ocaml and rescript and we find ourselves wanting to have different annotations in the two of them, like we often want to have list in ocaml and array in rescript, genType annotations in rescript and deriving in ocaml, etc. 

One solution would be to have a different tag name for rescript, something like `<rescript>` that would accept the same set of parameters as ocaml.

There are two ways to avoid it being a breaking change:
1. Have a new `-res` option for atdgen CLI that would ignore `<ocaml>` tags.
2. Make `<rescript>` tags override `<ocaml>` tags, but that would mean we would need to add missing values for ocaml, like `<ocaml repr="list">`.

I'd personally be in favor of 1., but I'm all ears for arguments if some of you prefer option 2. I don't know what's mélange maintainers preferences on that, like should they rather stick to `<ocaml>` tag, have their own, share one with rescript, etc.

Please chime in if you're rescript or mélange users that use ATD (👋 @TheSpyder @anmonteiro @wokalski and others I likely forget)

wokalski:
We discussed it internally. I have no preference. Either approach is good. We want to implement it ASAP so any input is appreciated. If you don't want to implement either, this information would be appreciated too.

TheSpyder:
I occasionally consider forking `bs-atdgen-codec-runtime` to experiment with making it more ReScript centric, if a ReScript mode is added to atd that’s probably a good time to do it 🤔

mjambon:
> 1. Have a new -res option for atdgen CLI that would ignore <ocaml> tags.

Instead of adding a command-line option, I would consider using a global annotation within the ATD file. Command line options should be reserved to deviate from the default behavior for the target language. The default behavior for the target languages should be encoded in the ATD file so that the recipient of an ATD file can build their code with the usual commands.

> 2. Make <rescript> tags override <ocaml> tags, but that would mean we would need to add missing values for ocaml, like <ocaml repr="list">.

This is what I would do to keep things simple.

[...]

PR #418, opened 2024-12-18
annotation filters with cli arguments

eWert-Online:
This implements the suggestion from @tsnobip and @mjambon in #324.

- [x] The command line can specify multiple tags (`-tag producer -tag res`)
- [x] The atd annotations can specify multiple tags (`<ocaml tag="producer" tag="foo" name="Foo">`)
- [x] An annotation that doesn't specify a tag is assumed to carry all the tags 
        (`<ocaml name="Foo">` is selected by `-tag bar` but `<ocaml tag="foo" name="Foo">` is not selected by `-tag bar`)

The tags are currently implemented by removing all attributes from the AST which do not match the cli argument(s).
This is the easiest way I could think of to implement that feature. If there is a better way or this leads to problems, please let me know 🙂 

I am not sure, if the schema changes I made, make sense. I just allowed `<... tag="foo">` fields in every attribute location (except for doc attributes).

### PR checklist

### DIFF
--- a/atd/src/annot.ml
+++ b/atd/src/annot.ml
@@ -214,3 +214,16 @@ let validate schema root =
   List.iter (fun sec ->
     validate_section sec root
   ) schema
+
+let filter_by_tags ~tags full_module =
+  let aux (annot : t) =
+    annot |> List.filter (fun (section, _) -> (
+      match tags, fields annot ~section ~field:"tag" with
+      | _, [] -> true
+      | tags, fields ->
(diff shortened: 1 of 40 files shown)


smondet (approved):

eWert-Online:
Hello 👋 

Is there anything I can do to get this moving forward? 

I still would like to get some feedback on the implementation 🙂

Also I don't know where / how to update the documentation.

Issue #416, opened 2024-12-10
group string_to string_of functions into record


ygrek:
this will help consumers to write generic code where this pair of functions can be passed around (it is ofc possible to do now just more verbose and annoying)

smondet:
Instead of a record why not a generate a submodule:

```
module Type_name = struct
  type t = type_name
  let of_string = type_name_of_string
  (* etc ...*)
end
```

- can still be used as a record (first class module)
    - in that case the modules are structurally typed so no need to share a
      record type with ATD libs from code using that feature
- can also be more easily used as a common functor argument

ygrek:
thanks!

PR #420, opened 2025-01-03
Add option `-j-gen-modules` to generate JSON generic submodules → #416

smondet:
### PR checklist

Cf. interesting bit in the tests:
https://github.com/ahrefs/atd/blob/923489d3175a8bd818e10639cfc45c67142604e9/atdgen/test/test_atdgen_main.ml?plain=1#L664-L676


atdgen/bin/ag_main.ml
```diff
@@ -184,6 +185,12 @@ let main () =
           refuse to print NaN and infinities (implying -json mode
           unless another mode is specified).";
 
+    "-j-gen-modules",
+    Arg.Unit (fun () ->
+                add_generic_modules := true),
+    "
+          Add genericity-friendly modules to -j output.";
```
mjambon, line 192: Can you specify what we should expect here and/or add it to the docs?

smondet: Added both, let me know what you think

mjambon (approved):
thanks!

PR #422, opened 2025-01-17
Add '--ignore' option to atddiff to ensure no types are accidentally unchecked

mjambon:
See the help message for an explanation of what the feature does.

We use this at Semgrep within a more complicated script to track that our APIs remain compatible from one release to another... and we do have multiple interfaces in a single ATD file, some of which are fine to break and others not. The triaging between types that need to be checked and the rest is done with the `--types` and now with this new `--ignore` option.

### PR checklist


aryx:
So what is the command to use in semgrep-interfaces to make sure that any backward incompatible change
detected from the root cli_output is reported, but anything that is backward incompatible detected from core_output
is not ? (and there is the subtelity that cli_output_extra is both referenced from cli_output and core_output, but 
the cli_output should take priority here so that it's not ignored even though it's referenced from core_output too).

mjambon:
> So what is the command to use in semgrep-interfaces to make sure that any backward incompatible change
> detected from the root cli_output is reported, but anything that is backward incompatible detected from core_output
> is not ? (and there is the subtelity that cli_output_extra is both referenced from cli_output and core_output, but
> the cli_output should take priority here so that it's not ignored even though it's referenced from core_output too).

It's what I put in the comment in the PR there:

```
    # We don't check for incompatibilities in the pysemgrep/semgrep-core
    # interface because these two programs are always distributed together,
    # allowing their interface to change freely.
    #
    # TODO: add the option '--ignore core_output,targets' so as to produce
    # an error if we fail to update the '--types' option when adding
    # future new type definitions. This requires atddiff > 2.15.
    atddiff_options="--no-locations --backward --types ci_scan_complete_response,ci_scan_results_response,deployment_response,diff_files,function_call,function_return,partial_scan_result,scan_config,scan_request,scan_response,tests_result"
```

So the command is:
```
$ atddiff --types ci_scan_complete_response,ci_scan_results_response,deployment_response,diff_files,function_call,function_return,partial_scan_result,scan_config,scan_request,scan_response,tests_result --ignore core_output,targets OLD NEW
```

aryx:
the list for --types and --ignore seems wrong.
I don't see cli_output for example, which should be in --types I guess, and
function_call and function_return should be in --ignore as they don't have to be backward compatible.

mjambon:
@aryx this conversation is for semgrep, not for ATD.

Issue #391, opened 2023-11-09
atdgen: menhir error


aryx:
With this file:
```
(* New Semgrep syntax (hence the v2) specified using ATD instead of jsonschema.
 *
 * For more information on the new syntax, see:
 *  - Brandon's community Slack post announcing the new syntax
 *    https://semgrep.slack.com/archives/C018NJRRCJ0/p1698430726062769?thread_ts=1698350734.415849&cid=C018NJRRCJ0
 *  - Brandon's slides
 *    https://docs.google.com/presentation/d/1zzmyFbfNlJqweyzuuFlo4zpSs3Gqhfi6FiNRONSEQ0E/edit#slide=id.g1eee710cdbf_0_26
 *  - Pieter's video
 *    https://www.youtube.com/watch?v=dZUPjFvknnI
 *  - Parsia's blog post
 *    https://parsiya.net/blog/2023-10-28-semgreps-experimental-rule-syntax/
 *
 * Note that even if most Semgrep users use YAML to write a rule, and not JSON,
 * we still use a JSON tool (here ATD, but also jsonschema) to specify
 * the rule schema because YAML is a superset of JSON and can be
 * mechanically translated into JSON; there is no yamlschema
 * (see https://json-schema-everywhere.github.io/yaml). To add even more
 * confusion, a jsonschema can actually be specified using YAML (like in
 * rule_shema_v1.yml), and so one can use YAML syntax to specify the
 * JSON schema of files actually written in YAML (hmmm).
 *
 * Jsonschema is powerful but also arguably complicated and so it
 * might be simpler for many Semgrep developers (and also some Semgrep
 * users) to use ATD to specify and understand the schema of a rule.
 * It could provide a better basis to think about future syntax extensions.
 *
 * This file is now also used for some rule validation in
 * `semgrep --validate --develop`.
 *
 * Note that this file does not replace Parse_rule.ml nor Rule.ml. We still
 * want to accept the old syntax in Parse_rule.ml and also parse with
 * position information and error recovery which ATD does not provide.
 * This files does not replace either (yet) rule_schema_v1.yml which is
 * more complete.
 *
 * TODO:
 *  - taint
 *  - extract
 *  - r2c-internal-project-depends-on-content
 *  - secrets
 *  - steps (and join?)
 *  - generalized taint
 *  - new metavariable types
 *  - new 'anywhere:'
 *)

(*****************************************************************************)
(* Basic types and string aliases *)
(*****************************************************************************)

[...]

PR #426, opened 2025-03-07
Catch and improve display of generic `Parser.Error` 

smondet:
Cf. #391

I tried adding to the menhir grammar but would get shift/reduce conflicts and/or would be incomplete.

Catching and improving the error message is what Menhir themselves do:
https://gitlab.inria.fr/fpottier/menhir/blob/master/src/stage2/Driver.ml

### PR checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@ Unreleased
 ----------
 
 * atdgen: Add option `-j-gen-modules` to generate JSON generic submodules (#420)
-
+* atd-parser: improve (syntax) error messages (#426)
 
 2.16.0 (2025-01-22)
 -------------------
--- a/atd/src/util.ml
+++ b/atd/src/util.ml
@@ -11,7 +11,19 @@ let read_lexbuf
     lexbuf =
 
   Lexer.init_fname lexbuf pos_fname pos_lnum;
-  let head, body = Parser.full_module Lexer.token lexbuf in
+  let head, body = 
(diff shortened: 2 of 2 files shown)


smondet:
```
 $ cat > /tmp/bug.atd <<EOF
type t = {
  type: string;
  world: float;
}
EOF
 $ dune exec atdgen/bin/ag_main.exe -- /tmp/bug.atd
Syntax error at File "/tmp/bug.atd", line 3, characters 2-6
```

smondet:
```
 $ cat > /tmp/bug.atd <<EOF
type t = {
  type: string;
  world: float;
}
EOF
 $ dune exec atdgen/bin/ag_main.exe -- /tmp/bug.atd
File "/tmp/bug.atd", line 3, characters 2-6: Syntax error
```

When Menhir does the job itself it look slightly different:

```
 $ cat > /tmp/bug.atd <<EOF
type x = int
type t = {
  different: case;
  type: string;
  world: float;
}
EOF
 $ dune exec atdgen/bin/ag_main.exe -- /tmp/bug.atd
File "/tmp/bug.atd", line 4, characters 2-6:
Expecting '}'
```

(`type` is not the first field here)

mjambon (approved):
Nice!

mjambon:
Note that if you write `Closes #391` on its own line in the PR description, GitHub will close the issue for you when this gets merged.

PR #434, opened 2025-10-01
Add automated ATD rule generation guide for Dune

octoio:
### PR checklist


Khady:
generally looking good, thanks. We have something similar internally, really useful. Could potentially be good to ship such an executable directly.

I wonder if the mode fallback is good or not. Might be a trap for people who are not very careful.

octoio:
> I wonder if the mode fallback is good or not. Might be a trap for people who are not very careful. But no strong opinion yet.

I can definitely update to remove the fallback mode.

PR #438, opened 2025-12-09
Bring back start_any_variant

kit-ty-kate:
As a demo for https://github.com/ocaml-community/yojson/pull/158#issuecomment-3631886133
The same thing needs to be done for `start_any_tuple` and friends but that's just a PoC.

Feel free to use this branch as see fit


mjambon:
I also came to the conclusion that importing just the ocamllex file (without the cppo directives) would be easiest. I'm trying this today.

kit-ty-kate:
Closing in favour of https://github.com/ahrefs/atd/pull/439 then

PR #498, opened 2026-04-16
atdml: support <ocaml attr=...> on record fields, variant constructors, and payloads

pedrobslisboa:
Add support for the ocaml attr annotation in three new positions in atdml: record fields, variant constructors, and variant payload types.

Type-definition-level attr (`[@@attr]`) was already supported; this extends it to the three remaining ATD positions.

Test plan:
- New variant attr snapshot test covering constructor-level and payload-level attr
- Existing attr snapshot test updated with a field-level attr example
- Docs updated in doc/atdml-reference.rst and CHANGES.md

🤖 Generated with Claude Code


mjambon:
Looks good to me, thanks! I fixed the merge conflict on the changelog.

mjambon:
dlang.org is down, giving us an error on these lines:
```
# To get latest ldc version
source $(curl https://dlang.org/install.sh | bash -s ldc -a)
echo "source $(~/dlang/install.sh ldc -a)" >> "$BASH_ENV"
```
Looking into using the ubuntu package instead (https://github.com/ahrefs/atd/pull/499)

mjambon:
Hmm. I'll let you rebase against the `master` branch. It should solve the `source: filename argument required` error.

Khady:
(I think that this was meant to be in atdgen, not atdml)

